PR #22, opened 2021-06-04
Generalization of ou-filter. [Refactoring]

Vitalii-Misechko:
This PR extracts `ou-filter`'s external interactions into `OuFilterDataManager` class. Basically, the class specifies API that should be implemented in order to use `ou-filter` component. The next step will be to move `ou-filter` to its own repo

FYI @anhill-D2L @hyehayes @johngwilkinson @NicholasHallman @bemailloux @JikaiLong


components/ou-filter/ou-filter.js
```diff
@@ -1,17 +1,65 @@
...
+	 * @returns {Object} - { PagingInfo:{ HasMoreItems: boolean, Bookmark: string }, Items: [ [ID, NAME, TYPE, PARENTS, ACTIVE_STATUS]] }. Fields indices are defined in tree-filter.js
+	 */
+	// eslint-disable-next-line no-unused-vars
+	async orgUnitSearch(searchString, bookmark) {
+		return null;
+	}
+
+	/**
+	 * Should access any mobx observable that is changed when orgUnitTree is updated.
+	 * orgUnitTree is not mobx-ified in order to cause direct mobx-subscription.
```
johngwilkinson, line 35: I think orgUnitTree does contain mobx observables, doesn't it? I'm not clear why this is needed.

Vitalii-Misechko: > I think orgUnitTree does contain mobx observables, doesn't it?

No, It doesn't. orgUnitTree is vanilla JS class. At some point, it uses `serverData` that does not contain observables as well (have a look at the comment below that describes why `serverData` does not have any observables)

```JS
		// because this._serverData itself is only updated onServerDataReload, we can safely use a simple
		// counter to let mobx know it has changed, rather than incurring the overhead of mobx-ifying
		// all the server data. This gives considerable speedup across the app for 50k enrollments.
		this._serverDataProxy = 0;
```

> I'm not clear why this is needed.

`div` is removed from `OuFilter` due to there is no CSS for `ou-filter` and `?loading` is not used anywhere as well.
But this `div` has side effect it causes mobx call `render` because `this.data.isLoading` is called in `render` method.
In order to preserve side effect but get rid of unused `div` I had to create `subscribeOnTreeDataChange`
```JS
render() {
...
		return html`<div class="ou-filter" ?loading="${this.data.isLoading}">
			<d2l-insights-tree-filter
```

I hope this clarify thing a bit

johngwilkinson: Isn't it a `Tree`? It's decorated here: https://github.com/Brightspace/insights-engagement-dashboard/blob/master/components/tree-filter.js#L440

Vitalii-Misechko: You're right, it's `Tree`, my bad. Probably, I should investigate further in order to figure out why removing `div` causes
this (picture below). By now, I see that Tree is updated but it's not rendered.

image

Vitalii-Misechko: Thank you for noticing it. I managed to get rid of this workaround by making `orgUnitTree` observable.

P.S. Basically, the reason was that mobx didn't know that  `orgUnitTree` changed

model/quizData.js
```diff
@@ -14,6 +16,37 @@ const defaultServerData = {
...
+ * Adapter class that represents all necessary API for OuFilter
+ */
+class QuizOuFilterDataManager extends OuFilterDataManager {
+
+	constructor(quizData) {
+		super();
+		this._quizData = quizData;
+	}
+
+	subscribeOnTreeDataChange() {
```
johngwilkinson, line 29: Can this method be removed now?

johngwilkinson (approved):

model/quizData.js
Vitalii-Misechko: Sure! Thanks.

bemailloux (approved):

PR #26, opened 2021-06-14
Quiz Table refactor

JikaiLong:
Refactored the table component to its own component. 

The table component takes 3 input parameters:
1. `tableData`: raw data that the table uses to display 
2. `tableConfig`:
  -`heading`: string or html template
  -`columnHeaders`: list of strings or templates
  -`displayData`: an string entry to the given data set, set it to null when directly passing array of data
  -`detaildataAttr`: list of objects - {`element`: attribute string or list of templates, `option`: 'template or null}
3. `skeleton` : boolean for skeleton template

Sample usage: 

```javascript

const serverData = new QuizData({ metronEndpoint: undefined });
serverData.loadData([6606, 6608], 1386);
const myTableConfig= {
		heading: 'quiz',
		columnHeaders: [
			'first column', 'second column'
		],
		displayData: 'quizzes',
		detaildataAttr: [
			{ element: 'averageScore', option: null },
			{ element: 'reliabilityCoefficient', option: null },
		]
	};

<assessment-quiz-table .tableData = ${serverData} .tableConfig = ${myTableConfig} ?skeleton = false ></assessment-quiz-table>

```
Functional Testing
All component: heading, column headers and table cells are unit tested. You can find the Unit test cases in test/components/assessment-quiz-table.

Performance
No difference, all visual diff tests passed.

Security
No change

Dev-Ops
NA

Cost
NA

UX
NA


NicholasHallman:
It's not immediately obviouse to me what the detailDataAttr is doing. Also does this PR add the styling configurations we need for rows and columns? I'm curiouse how we can dynamically color a data entry in a column using the API provided.

JikaiLong:
> It's not immediately obviouse to me what the detailDataAttr is doing. Also does this PR add the styling configurations we need for rows and columns? I'm curiouse how we can dynamically color a data entry in a column using the API provided.

The detailDataAttr is the data attribute that the table will be used to display. Say you want to display `serverData.quizzes ` then the detailDataAttr will be `'quizzes'`, it is an optional thing, if set to null, then `serverData.quizzes ` can passed directly into the table. 

data entry color is done in the parent class. 
For example, you could pass the follow template to the table using the tableConfig.heading
```
	const customHeader = html`
	<style>
	.custom-column-header {font-size: 50px; color: red }
	</style>
	<div class = 'custom-column-header' @click=${openclick} @keypress=${openclick}> custom column header</div>` ;
```

NicholasHallman:
> The detailDataAttr is the data attribute that the table will be used to display. Say you want to display serverData.quizzes then the detailDataAttr will be 'quizzes', it is an optional thing, if set to null, then serverData.quizzes can passed directly into the table. It could have a better name I suppose.

NicholasHallman:
> The detailDataAttr is the data attribute that the table will be used to display. Say you want to display serverData.quizzes then the detailDataAttr will be 'quizzes', it is an optional thing, if set to null, then serverData.quizzes can passed directly into the table. It could have a better name I suppose.

Personally, I think this is a little over complicated. Instead of telling the element how to pick data out of a large object we should be taking the required pieces out of the serverData and passing those into the table instead. I see that you did it this way so that the table isn't dependent on any specific object structure (which is great because it helps keep the component pure) but I think this is a round about way of acheiving that

> data entry color is done in the parent class.
For example, you could pass the follow template to the table using the tableConfig.heading

Your example for coloring the header works great because the header would get passed in as a tempalte. But how would we color the data elements? If the table only gets serverData and therefore raw values how do we tell it when we want those values to be styled or not?

JikaiLong:
> Personally, I think this is a little over complicated. Instead of telling the element how to pick data out of a large object we should be taking the required pieces out of the serverData and passing those into the table instead. I see that you did it this way so that the table isn't dependent on any specific object structure (which is great because it helps keep the component pure) but I think this is a round about way of acheiving that
>

Sorry, I miss read and thought you mean the displayData which is responsible for the the `serverData.quizzes` selection. The `detailDataAttr` is for template and data-entry passing. 

> Your example for coloring the header works great because the header would get passed in as a tempalte. But how would we color the data elements? If the table only gets serverData and therefore raw values how do we tell it when we want those values to be styled or not?
>

The template is actually a bit complicated. I was worrying about things like Quiz headers where they will take parameters like `id` and `name` and lots of local components to construct a single table cell. So instead, a list of component templates is created outside of the table, where they are passed in and the table is only responsible for rendering them. 
For example: we have this function `RenderComplexTemplate(inputdata)` which returns a very complex component for a single cell that has multiple functionalities like redirect to other pages.
then the parent should created a list like `[RenderComplexTemplate(data[0]), RenderComplexTemplate(data[1]), ....]` and passed to `detailDataAttr[0]` if this template is used for the first column.

NicholasHallman:
> Sorry, I miss read and thought you mean the displayData which is responsible for the the serverData.quizzes selection. The detailDataAttr is for template and data-entry passing.

In that case I think we should get rid of displayData

>  RenderComplexTemplate(inputdata) which returns a very complex component for a single cell that has multiple functionalities like redirect to other pages.

So in this case, we would pre-process serverData, iterating through the data and converting it into tempaltes for the table to use. In that case how do we style the td element that the table wraps? When we highlight a data point blue or red the whole table square is colored is that possible here?

components/table/assessment-quiz-table.js
```diff
@@ -0,0 +1,132 @@
...
+				${this._dataItems.map((item, index) => this._renderDetailsTableRow(item, index))}
+			</tbody>
+			</table>
+		</d2l-table-wrapper>
+		`;
+	}
+
+	_renderTableCell(data, dataAttr, index) {
+		if (dataAttr.option === 'template') {
+			return dataAttr.element[index];
```
NicholasHallman, line 107: I see in this case we expect the parent templating function's to return a TD Element

components/table/assessment-quiz-table.js
```diff
@@ -0,0 +1,132 @@
...
+
+			return html`<tr>
+				${Array(columnsCount).fill('').map(() => html`<td>
+					<div class='d2l-skeletize d2l-skeletize-95 d2l-body-standard'>${this.localize('dashboard:loading')}</div>
+				</td>`)}
+			</tr>`;
+		}
+		return html`
+			<tr>
+			${this.tableConfig.detaildataAttr.map(dataAttr => this._renderTableCell(item, dataAttr, index))}
```
NicholasHallman, line 125: Maybe a better name for detailedDataAttr would be rowOptions? Makes it a little easier to understand what it is working on

JikaiLong:
> In that case I think we should get rid of displayData
Yes, that makes sense. Will remove it. 

> So in this case, we would pre-process serverData, iterating through the data and converting it into tempaltes for the table to use. In that case how do we style the td element that the table wraps? When we highlight a data point blue or red the whole table square is colored is that possible here?
The parent actually creates and passes the `<td></td>` element into the table where they can style the element outside of the table. 
So you mean the `<td></td> `element shouldn't get passed from the parent? I could make a separated CSS config attribute to the cell but that means config structure will be more complex.

components/table/assessment-quiz-table.js
JikaiLong: Yes.

components/table/assessment-quiz-table.js
JikaiLong: Sounds good, I was trying to figure out a better name but had nothing in mind.

JikaiLong: Actually it is not ROW options, it is used to target columns, so columnOptions seems better

NicholasHallman:
> So you mean the `<td></td>` element shouldn't get passed from the parent? I could make a separated CSS config attribute to the cell but that means config structure will be more complex.

Yeah I don't think we should rely on the developer who is using the table to pass td elements in if they want to change the styling. I'm under the impression that the table should supply the structure and default styles and passing in your own structrue could ruin the accessbility of the table.

JikaiLong:
> Yeah I don't think we should rely on the developer who is using the table to pass td elements in if they want to change the styling. I'm under the impression that the table should supply the structure and default styles and passing in your own structrue could ruin the accessbility of the table.
>
I could add a new config to add style to td but it should specifically tells which cell needs to be styled.

NicholasHallman:
I think we should supply functions that can run over the data in a row or column and spit out a styleMap and the formatted data given some set of options.

i.e

```javascript
render() {
  // run by the table to dynamically determine the styling of a piece of data
  const formatColumn1 = (data) => {
    const formtedValue = data + '%'; // format value as a percent
    const style = styleMap({
      'background-color': data > 10 ? 'blue' : 'red'
    });
    return { value: formatedValue, style }
  }
  const column1Template = this.tableHeaderWithTooltip('column 1'); // format the header into a template with tooltips
  const options = {
    headers: [column1Template, column2Template],
    columnFormat: [ formatColumn1, formatColumn2 ],
  }

  return html`<assessment-quiz-table data=${data} options=${options}></assessment-quiz-table>`
}
```

JikaiLong:
> I think we should supply functions that can run over the data in a row or column and spit out a styleMap and the formatted data given some set of options.
> 
Sounds good to me. In the mapping function, the data will be a data entry for the entire row since one custom cell may take multiple attributes from that row.

NicholasHallman:
>  the data will be a data entry for the entire row since one custom cell may take multiple attributes from that row.

I'm not sure I understand what you mean. I was thinking that as the table rendered each cell it would take the formatting function for that column and run it over the data to go into that cell. i.e

``` javascript
// inside assessment-quality-table
function _renderCell(data, formatFunction) {
  let { value, style } = formatFunction(data);
  return html`<td style=${style}> ${value} </td>`;
}

function _renderRow(rowNum) {
  let cells = []; 
  let lengthOfRow = this.options.headers.length;
  for (let i = 0; i < lengthOfRow; i++){
    cells.push(this._renderCell( this.data[rowNum][i], this.options.columnFormat[i] ));
  }
  return html`<tr>${cells}</tr>`
}
```

If we want multiple kinds of attributes we can encode that into the data. For example an element that includes a link could look like

```javascript
// data is the text, and the href in an array
// example: [ 'home', '/' ]
const formatFunction(data) {
  return { value: html`<a href=${data[1]}>${data[0]</a>}`
}
```

JikaiLong:
> I'm not sure I understand what you mean. I was thinking that as the table rendered each cell it would take the formatting function for that column and run it over the data to go into that cell. i.e
> 
Yes. I basically meant that.

> If we want multiple kinds of attributes we can encode that into the data. For example an element that includes a link could look like
> 
Right, I am going to do something similar to the quizHeaders.

Vitalii-Misechko:
Just a reminder that it's important to add all team members into FYI for each PR review

> FYI @anhill-D2L @hyehayes @johngwilkinson @NicholasHallman @bemailloux @JikaiLong @SEsseltine

components/table/assessment-quiz-table.js
```diff
@@ -0,0 +1,132 @@
...
+-columnHeaders: list of strings or templates
+-displayData: an string entry to the given data set, set it to null when directly passing array of data
+-detaildataAttr: list of objects - {element: attribute string or list of templates, option: 'template or null}
+*/
+
+class AssessmentTable extends SkeletonMixin(Localizer(MobxLitElement)) {
+
+	static get properties() {
+		return {
+			isDemo: { type: Boolean, attribute: 'demo' },
```
Vitalii-Misechko, line 21: I believe we can get rid of `isDemo` inside the table component.

components/table/assessment-quiz-table.js
```diff
@@ -0,0 +1,132 @@
...
+		`];
+	}
+	constructor() {
+		super();
+		this.isDemo = false;
+	}
+
+	render() {
+		return html`
+		${this._renderQuizDetailsTable()}
```
Vitalii-Misechko, line 64: `minor`:  we can inline all the code right here. There are no benefits of the existence of `_renderQuizDetailsTable`

components/table/assessment-quiz-table.js
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	get _dataItems() {
+		if (this.skeleton) {
+			return Array(DEFAULT_PAGE_SIZE).fill({});
+		}
+		if (this.tableConfig.displayData !== null) {
+			return this.tableData[this.tableConfig.displayData];
+		}
+		return this.tableData;
```
Vitalii-Misechko, line 75: Looks a bit complicated. I believe we can make a more simple code design. It's up to a developer to provide and format the necessary data for the component. Therefore, a single property for data should be enough.

components/table/assessment-quiz-table.js
```diff
@@ -0,0 +1,132 @@
...
+	get _columnHeaders() {
+		return this.tableConfig.columnHeaders;
+	}
+
+	_renderallColumnHeaders() {
+		return html`			
+		<thead>
+		<th>${this._columnHeaders[0]}</th>
+		${this._columnHeaders.slice(1).map(column => html`<th>${column}</th>`)}
+	</thead>
```
Vitalii-Misechko, line 87: Using magic numbers is [a code smell.](https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad) Try to avoid them

Vitalii-Misechko: `nit`: could you also fix formatting

components/table/assessment-quiz-table.js
JikaiLong: Yes. Nick pointed it out as well. I already removed it.

components/table/assessment-quiz-table.js
JikaiLong: The problem is that if I do not use `<th>${this._columnHeaders[0]}</th>` to initialize the first cell, the header structure is not displayed for some reason.

pages/assessment-quality-dashboard-view.js
```diff
@@ -20,13 +20,28 @@ class AssessmentQualityDashboardView extends SkeletonMixin(Localizer(MobxLitElem
...
+			],
+			displayData: 'quizzes',
+			detaildataAttr: [
+				{ element: '', option: 'template' },
+				{ element: 'averageScore', option: null },
+				{ element: 'reliabilityCoefficient', option: null },
+				{ element: 'numberOfQuizItems', option: null },
+				{ element: 'flaggedQuizItems', option: null },
+				{ element: 'numberOfLearnerAnswers', option: null }
+			]
```
Vitalii-Misechko, line 43: I see two types of elements in `detaildataAttr`:
1. `{ element: string: option: null }`
2. `{ element: string: opton: 'template' }` - special case type

I would suggest splitting these types. The special case type can be a separate property/array

pages/assessment-quality-dashboard-view.js
```diff
@@ -165,90 +129,18 @@ class AssessmentQualityDashboardView extends SkeletonMixin(Localizer(MobxLitElem
...
-			</tr>`;
-		}
-
-		const courseNameText = this.localize('dashboard:quizDetailsCourseNameTemplate', { courseName: quiz.courseName, courseId: quiz.courseId });
-		const courseNameTooltipId = `course-tooltip-${quiz.quizId}`;
-		const quizId = `quiz-tooltip-${quiz.quizId}`;
+	_renderQuizTable() {
+		const template = this._table_quiz_name_header_template_list(this.serverData.quizzes);
+		this.tableConfig.columnHeaders = this._d2l_table_column_data();
+		this.tableConfig.detaildataAttr[0].element = template;
```
Vitalii-Misechko, line 140: `minor` 
1. AFAIK, we don't use snake_case. 
2. `template` is not reused. no need to declare this const
3. magic numbers is a code smell

pages/assessment-quality-dashboard-view.js
```diff
@@ -273,6 +165,118 @@ class AssessmentQualityDashboardView extends SkeletonMixin(Localizer(MobxLitElem
 		event.stopPropagation();
 		this.serverData.selectedOrgUnitIds = event.target.selected;
 	}
+
+	// the following is for table template creation targetting AQ dashbaord
+
+	_renderHeaderWithToolTip(uniqueId, headerLangTerm, tooltipTextLangTerm) {
```
Vitalii-Misechko, line 171: this code should be moved into table componenent. `tooltipId` can be generated as `d2l-table-tooltip-${columnNumber}-${rowId}`

pages/assessment-quality-dashboard-view.js
```diff
@@ -273,6 +165,118 @@ class AssessmentQualityDashboardView extends SkeletonMixin(Localizer(MobxLitElem
...
+		return template_list;
+	}
+
+	_redirectToQuiz(e) {
+		const preId = 'quiz-tooltip-';
+		const quizId = e.target.id.substring(preId.length);
+		redirect(`/quiz/${quizId}`);
+	}
+
+	_d2l_table_quiz_name_header(data) {
```
Vitalii-Misechko, line 221: I assume the code inside of this method will be repeated as many times as the number of rows. If so we need to reconsider this solution. You might be interested in how this was solved in [d2l-insights-table](https://github.com/Brightspace/insights-engagement-dashboard/blob/06ad425e7ca502c89e59cff85427e9b61a86aaf5/components/table.js)

JikaiLong: I just checked the rendercell function:
https://github.com/Brightspace/insights-engagement-dashboard/blob/06ad425e7ca502c89e59cff85427e9b61a86aaf5/components/table.js#L216 
and the code is also getting called multiple times from this call `${this.columnInfo.map(this._renderHeaderCell, this)}`, so I wonder what would be a better fix for this?

pages/assessment-quality-dashboard-view.js
JikaiLong: But then the parent is required to pass `uniqueId, headerLangTerm, tooltipTextLangTerm` into the table component for it to render. And the header component also takes templates as input, so passing these parameters into the table component seems a bit duplicate I think.

Vitalii-Misechko: > But then the parent is required to pass uniqueId, headerLangTerm, tooltipTextLangTerm into the table component for it to render. 

* Component design. Essentially it's natural/idiomatic to me that the code provides columns' descriptions for the table component. A column description can contain (but not limited to) column name, type, column header/value formatter, tooltip text etc. 
* Regarding `uniqueId`. It rather unique **column** id. In this case, column index or column name is totally fine.

> And the header component also takes templates as input, so passing these parameters into the table component seems a bit duplicate I think.

* Component design. IMHO, the code should not provide a header tooltip template as a parameter of the table component because there is a simpler implementation when `_renderHeaderWithToolTip` is a part of the table component.  

```JS
// alternative implementation of
_renderallColumnHeaders()
	return html`			
		<thead>
			${this.columnHeaders.map(
			(header, idx) =>  header.tooltipTextLangTerm
			  ? this._renderHeaderWithToolTip(idx, header.headerLangTerm, header.tooltipTextLangTerm) 
			  : html`<th>${this.localize(header.headerLangTerm)}</th>`)
		</thead>`
```

pages/assessment-quality-dashboard-view.js
Vitalii-Misechko: Sorry, I didn't bring enough context. I meant to say that I suspect that this code includes `<script>` HTML many times in the component output which is not good. We could introduce rendering similar to `d2l-insights-table` [implementation](https://github.com/Brightspace/insights-engagement-dashboard/blob/06ad425e7ca502c89e59cff85427e9b61a86aaf5/components/table.js#L317) and move this code inside the table component avoiding excessive code generation

```JS
if (columnType === COLUMN_TYPES.ROW_SELECTOR) {
/// One type of rendering
} else if (columnType === COLUMN_TYPES.TEXT_SUB_TEXT) {
...
} else {
// general type of cell rendering
}
```

components/table/assessment-quiz-table.js
JikaiLong: I am still stuck on fixing this one. Apparently using a loop(map) to create component is not going to generate the entry point tag for the first row `d2l-table-cell-first` and first cell `d2-table-cell-first`, therefore the table format will not applied to any of those headers.  However, the first cell must be defined manually not in a loop to have a tag on its creation.

JikaiLong: update: fix them by auto assign class tags similar to engagement dashboard

JikaiLong:
Update:
The new config file now looks like  this:
```
this.tableConfig = {
	heading: '',
	columnHeaders: [
		{ headerLangTerm: 'dashboard:quizDetailsQuiz', tooltipTextLangTerm: null },
		{ headerLangTerm: 'dashboard:quizDetailsAverageScore', tooltipTextLangTerm: null },
		{ headerLangTerm:  'dashboard:quizDetailsReliabilityCoefficient', tooltipTextLangTerm: 'dashboard:quizDetailsReliabilityCoefficientTooltip' },
	],
	columnCellValues:[
		null,
		'averageScore',
		'reliabilityCoefficient',
	],
	columnCellFormat: [
		format_d2l_table_quiz_name_header,
		null,
		null,
	],
};
```
where `format_d2l_table_quiz_name_header` is a function that takes data of a row and render a cell component, it returns a html template for content render and a `styleMap ` which is used to customize the cell itself: such as background color of the cell.

NicholasHallman:
What does `columnCellValues` do ?

JikaiLong:
> What does `columnCellValues` do ?

They are attributes of the data to be displayed. like quizzes.averageScore.

Vitalii-Misechko:
This is unfortunate, but it seems my merge to master created merge conflict here

JikaiLong:
> This is unfortunate, but it seems my merge to master created merge conflict here

Yea. I think I could either move the pagination part into the table or keep it outside in the dashboard view page. What do you think?

Vitalii-Misechko:
> > This is unfortunate, but it seems my merge to master created merge conflict here
> 
> Yea. I think I could either move the pagination part into the table or keep it outside in the dashboard view page. What do you think?

Let's keep it outside for simplicity. We can make another refactoring later if we change our mind

JikaiLong:
Conflict resolved. Now table should work with pagination.

pages/assessment-quality-dashboard-view.js
```diff
@@ -23,13 +25,98 @@ class AssessmentQualityDashboardView extends SkeletonMixin(Localizer(MobxLitElem
...
 
 	constructor() {
 		super();
 		this.isDemo = false;
+		const redirectToQuiz = (e) => {
+			const preId = 'quiz-tooltip-';
+			const quizId = e.target.id.substring(preId.length);
+			redirect(`/quiz/${quizId}`);
+		};
+		const format_d2l_table_quiz_name_header = (data) => {
```
Vitalii-Misechko, line 41: The main question is what kind of component we are trying to build here? If this is some kind of general table component and it is going to be used D2L -wide then this approach with external value formatter looks good, but the component design should be discussed with Dave Lockhard. If this is an application-wide component that is used only in AQ Dashboard then this is a bit overkill.

I assume we're building an application-wide component. Therefore, this code should be a part of the table component.

```JS
	_renderTableCell(row_item, attr, column_index) {
	        const columnType = this.tableConfig.columnType[column_index]

		if (columnType === ColumnTypes.CLICKABLE_QUIZ_NAME) {
			return this.format_d2l_table_quiz_name_header(row_item)`;
		}

		return html`<td>${row_item[attr]}</td>`;
	}
```

Vitalii-Misechko: `nit` Could you go over your PR and 

1. replace `snake_case `with `camelCase`? `SCREAMING_SNAKE_CASE` is used for constants.
2. remove space in attribute assignments for html code. For instantce, rplace this `<tr class = 'd2l-class-name'>` with this `<tr class='d2l-class-name'>`

pages/assessment-quality-dashboard-view.js
```diff
@@ -23,13 +25,98 @@ class AssessmentQualityDashboardView extends SkeletonMixin(Localizer(MobxLitElem
...
+			heading: '',
+			columnHeaders: [
+				{ headerLangTerm: 'dashboard:quizDetailsQuiz', tooltipTextLangTerm: null },
+				{ headerLangTerm: 'dashboard:quizDetailsAverageScore', tooltipTextLangTerm: null },
+				{ headerLangTerm:  'dashboard:quizDetailsReliabilityCoefficient', tooltipTextLangTerm: 'dashboard:quizDetailsReliabilityCoefficientTooltip' },
+				{ headerLangTerm: 'dashboard:quizDetailsNumberOfQuizItems', tooltipTextLangTerm: null },
+				{ headerLangTerm:  'dashboard:quizDetailsFlaggedQuizItems', tooltipTextLangTerm: 'dashboard:quizDetailsFlaggedQuizItemsTooltip' },
+				{ headerLangTerm: 'dashboard:quizDetailsNumberOfLearnerAnswers', tooltipTextLangTerm: null },
+			],
+			columnCellValues:[
```
Vitalii-Misechko, line 103: `nit` can we rename `columnCellValues` to `columnValuePropertyNames` or `columnValueGetters`

Vitalii-Misechko (changes requested):

pages/assessment-quality-dashboard-view.js
NicholasHallman: The reason I suggested the formatters was because this table has a bunch of different ways of displaying data and we this PR doesn't pre-process the data. Theres two ways to do this

1. We iterate through the data, and turn it into the format we want pre-emptively (like we do for highcharts and in engagement)
2. We give lambdas that format to a specification.

The ways we format data in the table are
- Cells can be colored red
- That little red posfix dot
- Links, tooltips, etc

So we either do it before, of during render. TBH Before is probably better. Since we can use mobx to compute it and pass in a processed data array for the table to draw and not the serverdata object

pages/assessment-quality-dashboard-view.js
JikaiLong: columnValuePropertyNames sounds good.

components/table/assessment-quiz-table.js
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	_renderallColumnHeaders() {
+		return html`			
+		<thead>
+			<tr class = 'd2l-table-row-first'>
+				${this._columnHeaders.map(
+		(header, idx) =>  (header.tooltipTextLangTerm
+			? this._renderHeaderWithToolTip(idx, header.headerLangTerm, header.tooltipTextLangTerm)
+			: html`<th class = ${ idx === 0 ? 'd2l-table-cell-first' : null }>${this.localize(header.headerLangTerm)}</th>`))}
```
Vitalii-Misechko, line 89: `nit` Have you considered using `classMap`? Here is an [example](https://github.com/Brightspace/insights-engagement-dashboard/blob/06ad425e7ca502c89e59cff85427e9b61a86aaf5/components/table.js#L296)

JikaiLong: I see. classMap is definitely better. Thanks for letting me know.

pages/assessment-quality-dashboard-view.js
JikaiLong: Just a bit curious: the column type in the engagement dashboard has the following properties:

export const COLUMN_TYPES = {	ROW_SELECTOR: -1,	NORMAL_TEXT: 0,	TEXT_SUB_TEXT: 1,	SUB_COLUMNS: 2};<br class="Apple-interchange-newline">

Should I create my own property or use one of these?

Vitalii-Misechko: > The reason I suggested the formatters was because this table has a bunch of different ways of displaying data and we this PR doesn't pre-process the data. Theres two ways to do this

I see. Lambdas are great and powerful and can reduce the complexity of the component. If you guys feel that this is the right design for this component I don't have strong objections against it. If you hesitate I would suggest 'selling' this design idea to John or Bryan. If you succeed you will be sure that it's optimal in our team.

NicholasHallman: Hmm it's a little more complicated then I had thought since the api has changed a lot between these two projects...

Vitalii-Misechko: > Just a bit curious: the column type in the engagement dashboard has the following properties:
> `ROW_SELECTOR: -1, NORMAL_TEXT: 0, TEXT_SUB_TEXT: 1, SUB_COLUMNS: 2`
> Should I create my own property or use one of these?

These are good examples but I believe we will end up with a bit different types.

JikaiLong: So here are two options:

1. Create a list of html cell templates using our current data, and pass them into the table then let the table iterate through each of them to display
2. Create a lambda function and pass it into the table, the table takes the function and input data, then use this function to render html cell templates and display

I asked @bemailloux  about two different ways of design and here is his response:

"There's a more rigid approach you could consider (which is basically (1) but possibly a bit more organized / less flexible). For every table cell type you want to consider, create a LitElement component for it. Then have the table component decide which table cell component to use for each column.

So I'd consider (1) or my alternative. I'm not a fan of (2) because it just doesn't really fit the "components" paradigm. If you're going to pass in a rendering lambda for each column type, you may as well just turn it into a component."

What do you guys think? @Vitalii-Misechko  @NicholasHallman

Vitalii-Misechko: I would love to see implemented option (1). But, I'm fine with any of your choice.

JikaiLong:
Update:
All template are passed in as lambdas.
New config file looks like this:

```
const noStyle = styleMap();
const averageScoreTemplate = (data) => { return { value: html`<div>${data.averageScore}</div>`, noStyle };};
const tableConfig= {
		heading: 'quiz',
		columnHeaders: [
			{ headerLangTerm: 'dashboard:quizDetailsQuiz', tooltipTextLangTerm: null }
		],
		columnCellFormat: [averageScoreTemplate]
};
```
Test Cases Updated. 
Code styles and minor things are fixed.

components/table/assessment-quiz-table.js
```diff
@@ -0,0 +1,140 @@
...
+	_renderallColumnHeaders() {
+		const styleFirstRow = { 'd2l-table-row-first': true };
+		const styleFirstCell = { 'd2l-table-cell-first': true };
+		return html`			
+		<thead>
+			<tr class=${classMap(styleFirstRow)}>
+				${this._columnHeaders.map(
+		(header, idx) =>  (header.tooltipTextLangTerm
+			? this._renderHeaderWithToolTip(idx, header.headerLangTerm, header.tooltipTextLangTerm)
+			: html`<th class=${ idx === 0 ? classMap(styleFirstCell) : '' }>${this.localize(header.headerLangTerm)}</th>`))}
```
Vitalii-Misechko, line 92: Let me suggest one minor improvement 
```JS
                const styleFirstCell = (idx) => ({ 'd2l-table-cell-first': idx === 0 });
                ....
				: html`<th class=${classMap(styleFirstCell(idx))}>${this.localize(header.headerLangTerm)}</th>`))}


```

Vitalii-Misechko: I didn't check the code so it might not work ;)

components/table/assessment-quiz-table.js
```diff
@@ -0,0 +1,140 @@
...
+	get _columnHeaders() {
+		return this.tableConfig.columnHeaders;
+	}
+
+	_renderallColumnHeaders() {
+		const styleFirstRow = { 'd2l-table-row-first': true };
+		const styleFirstCell = { 'd2l-table-cell-first': true };
+		return html`			
+		<thead>
+			<tr class=${classMap(styleFirstRow)}>
```
Vitalii-Misechko, line 88: `minor` this line `<tr class=${classMap(styleFirstRow)}>` does not have any benefits over simpler version `<tr class='d2l-table-cell-first'>`. So I'd suggest using simple version

components/table/assessment-quiz-table.js
```diff
@@ -0,0 +1,140 @@
...
+				${Array(columnsCount).fill('').map(() => html`<td>
+					<div class='d2l-skeletize d2l-skeletize-95 d2l-body-standard'>${this.localize('dashboard:loading')}</div>
+				</td>`)}
+			</tr>`;
+		}
+		return html`
+			<tr>
+				${this.tableConfig.columnCellFormat.map(formatFunction =>
+		this._renderTableCell(rowItem, formatFunction)
+	)}
```
Vitalii-Misechko, line 132: `nit` inconsistent indentation

pages/assessment-quality-dashboard-view.js
```diff
@@ -194,102 +147,105 @@ class AssessmentQualityDashboardView extends SkeletonMixin(Localizer(MobxLitElem
...
+		const redirectToQuiz = (e) => {
+			const preId = 'quiz-tooltip-';
+			const quizId = e.target.id.substring(preId.length);
+			redirect(`/quiz/${quizId}`);
+		};
+		const formatD2lTableQuizNameHeader = (data) => {
+			const courseNameText = this.localize('dashboard:quizDetailsCourseNameTemplate', { courseName: data.courseName, courseId: data.courseId });
+			const courseNameTooltipId = `course-tooltip-${data.quizId}`;
+			const quizId = `quiz-tooltip-${data.quizId}`;
+			const formatedValue =  html`
```
Vitalii-Misechko, line 166: `Suggestion`. What if we add `tableConfig.sharedStyles` and pass all styles below only once in order to avoid multiple inclusions of this code (code between script tags) into the component output. We could render these styles like the following code.
```JS
 	render() {
		return html`
		${this.tableConfig.sharedStyles}
		<h2 class="d2l-heading-3">${this.tableConfig.heading}</h2>
```

pages/assessment-quality-dashboard-view.js
```diff
@@ -194,102 +147,105 @@ class AssessmentQualityDashboardView extends SkeletonMixin(Localizer(MobxLitElem
...
+			<assessment-quiz-table .tableData = ${this._displayData} .tableConfig = ${tableConfig} ?skeleton = ${this.skeleton} ></assessment-quiz-table>
+			<d2l-labs-pagination
+			show-item-count-select
+			item-count-options="[10,20,50,100]"
+			page-number="${this._currentPage}"
+			max-page-number="${this._maxPages}"
+			selected-count-option="${this._pageSize}"
+			@pagination-page-change="${this._handlePageChange}"
+			@pagination-item-counter-change="${this._handlePageSizeChange}"
+			></d2l-labs-pagination>
```
Vitalii-Misechko, line 247: `nit` inconsistent formatting:
* extra spaces for attribute assignemts for `assessment-quiz-table`
* it seems inconsisten indentation for `d2l-labs-pagination`

```JS
// proper indentation
<d2l-labs-pagination
  show-item-count-select
  item-count-options="[10,20,50,100]"
```

test/components/assessment-quiz-table/assessment-quiz-table.test.js
```diff
@@ -0,0 +1,233 @@
+import '../../../components/table/assessment-quiz-table';
+import { expect, fixture, html } from '@open-wc/testing';
+import { QuizData } from '../../../model/quizData';
+import { runConstructor } from '@brightspace-ui/core/tools/constructor-test-helper.js';
+import sinon from 'sinon/pkg/sinon-esm.js';
+import { styleMap } from 'lit-html/directives/style-map';
+
+const serverData = new QuizData({ metronEndpoint: undefined });
+serverData.loadData([6606, 6608], 1386);
```
Vitalii-Misechko, line 9: `nit` you basically can write `serverData.loadData({})` now. These test parameters `[6606, 6608], 1386` don't do anythynig anymore

pages/assessment-quality-dashboard-view.js
JikaiLong: Totally agree.

test/components/assessment-quiz-table/assessment-quiz-table.test.js
JikaiLong: Thanks for letting me know. :)

pages/assessment-quality-dashboard-view.js
```diff
@@ -194,102 +147,105 @@ class AssessmentQualityDashboardView extends SkeletonMixin(Localizer(MobxLitElem
...
 					</div>
-				</td>
-				<td>${quiz.averageScore}</td>
-				<td>${quiz.reliabilityCoefficient}</td>
-				<td>${quiz.numberOfQuizItems}</td>
-				<td>${quiz.flaggedQuizItems}</td>
-				<td>${quiz.numberOfLearnerAnswers}</td>
-			</tr>
+			</div>`;
+			return { value: formatedValue, noStyle };
```
Vitalii-Misechko, line 211: `nit` Interesting. The rendering function tries to get `style` property but it's `undefined` which is okay/intended. Despite I like this code readability it doesn't make sense for rendering function due to it doesn't read `noStyle` property.

Vitalii-Misechko: I guess you wanted write something like this 
```JS
{ value: formatedValue, style: noStyle };
```

test/components/assessment-quiz-table/assessment-quiz-table.test.js
```diff
@@ -0,0 +1,233 @@
...
+		});
+	});
+
+	describe('render', () => {
+		it('should show assessment-quiz-table', async() => {
+			const el = await fixture(html`<assessment-quiz-table
+            demo
+            .tableData = ${serverData.quizzes}
+            .tableConfig = ${tableConfigOneColumn}
+        ></assessment-quiz-table>`);
```
Vitalii-Misechko, line 90: `nit` inconsistent indentation

test/components/assessment-quiz-table/assessment-quiz-table.test.js
```diff
@@ -0,0 +1,233 @@
...
+	});
+
+	describe('render default first column header', () => {
+		it('should show table first column header', async() => {
+			const el = await fixture(html`<assessment-quiz-table
+            demo
+            .tableData = ${serverData.quizzes}
+            .tableConfig = ${tableConfigOneColumn}
+        ></assessment-quiz-table>`);
+			const table = el.renderRoot.querySelector('d2l-table-wrapper').querySelector('table');
```
Vitalii-Misechko, line 105: `code improvement suggestion` I assume this CSS selector should work
```JS
el.renderRoot.querySelector('d2l-table-wrapper table')
```

JikaiLong: cool. I never tried that way. Will see if that works

test/components/assessment-quiz-table/assessment-quiz-table.test.js
```diff
@@ -0,0 +1,233 @@
...
+
+	describe('render default second column header', () => {
+		it('should show table second column header', async() => {
+			const el = await fixture(html`<assessment-quiz-table
+            demo
+            .tableData = ${serverData.quizzes}
+            .tableConfig = ${tableConfigTwoColumn}
+        ></assessment-quiz-table>`);
+			const table = el.renderRoot.querySelector('d2l-table-wrapper').querySelector('table');
+			const columnHeaders = table.querySelector('thead').querySelector('tr').querySelectorAll ('th');
```
Vitalii-Misechko, line 119: `same` inconsistent style, possible simple selector, extra spaces for attribute assignments

test/components/assessment-quiz-table/assessment-quiz-table.test.js
```diff
@@ -0,0 +1,233 @@
...
+
+	describe('render first column header with tool-tip', () => {
+		it('should show table first tooltip column header', async() => {
+			const el = await fixture(html`<assessment-quiz-table
+            demo
+            .tableData = ${serverData.quizzes}
+            .tableConfig = ${tableConfigTooltipColumn}
+        ></assessment-quiz-table>`);
+			const table = el.renderRoot.querySelector('d2l-table-wrapper').querySelector('table');
+			const columnHeader = table.querySelector('thead').querySelector('tr').querySelector('th').querySelector('div');
```
Vitalii-Misechko, line 132: `same` inconsistent style, possible simple selector, extra spaces for attribute assignments

test/components/assessment-quiz-table/assessment-quiz-table.test.js
```diff
@@ -0,0 +1,233 @@
...
+			const el = await fixture(html`<assessment-quiz-table
+            demo
+            .tableData = ${serverData.quizzes}
+            .tableConfig = ${tableConfigTwoColumn}
+        ></assessment-quiz-table>`);
+			const table = el.renderRoot.querySelector('d2l-table-wrapper').querySelector('table');
+			const firstRow = table.querySelector('tbody').querySelector('tr');
+			const cells = firstRow.querySelectorAll('td');
+			expect(cells[0].querySelector('div').innerText).equal('0.29');
+			expect(cells[1].querySelector('div').innerText).equal('0.19');
```
Vitalii-Misechko, line 149: `same` inconsistent style, possible simple selector, extra spaces for attribute assignments

test/components/assessment-quiz-table/assessment-quiz-table.test.js
```diff
@@ -0,0 +1,233 @@
...
+            demo
+            .tableData = ${serverData.quizzes}
+            .tableConfig = ${tableConfigTwoColumn}
+        ></assessment-quiz-table>`);
+			const table = el.renderRoot.querySelector('d2l-table-wrapper').querySelector('table');
+			const secondRow = table.querySelector('tbody').querySelectorAll('tr')[1];
+			const cells = secondRow.querySelectorAll('td');
+			expect(cells[0].querySelector('div').innerText).equal('0.5');
+			expect(cells[1].querySelector('div').innerText).equal('0.89');
+		});
```
Vitalii-Misechko, line 165: `same` inconsistent style, possible simple selector, extra spaces for attribute assignments. I'd suggest going over the whole file to adjust formatting

test/components/assessment-quiz-table/assessment-quiz-table.test.js
```diff
@@ -0,0 +1,233 @@
...
+		let originalOpen;
+		beforeEach(() => {
+			originalOpen = window.open;
+			Object.defineProperty(window, 'open', { value: sinon.stub(), configurable:true, enumerable:true, writable:true });
+		});
+
+		after(() => {
+			Object.defineProperty(window, 'open', { value: originalOpen, configurable:true, enumerable:true, writable:true });
+		});
+		it('should show table custom cells', async() => {
```
Vitalii-Misechko, line 178: missed CRLF before `it`

test/components/assessment-quiz-table/assessment-quiz-table.test.js
```diff
@@ -0,0 +1,233 @@
...
+			expect(firstCustomCell.innerText).equal('0.29');
+			const firstCustomCellStyle = window.getComputedStyle(firstCustomCell);
+			expect(firstCustomCellStyle.getPropertyValue('color')).equal('rgb(255, 0, 0)');
+			expect(firstCustomCellStyle.getPropertyValue('font-size')).equal('50px');
+			expect(secondCustomCell.innerText).equal('0.5');
+			const secondCustomCellStyle = window.getComputedStyle(secondCustomCell);
+			expect(secondCustomCellStyle.getPropertyValue('color')).equal('rgb(255, 0, 0)');
+			expect(secondCustomCellStyle.getPropertyValue('font-size')).equal('50px');
+		});
+		it('should open a new page when click on a custom cell', async() => {
```
Vitalii-Misechko, line 196: missed CRLF before `it`

test/components/assessment-quiz-table/assessment-quiz-table.test.js
JikaiLong: True. I will fix the whole style things for the test file.

Vitalii-Misechko (commented):
PR looks much better! Thank you @JikaiLong for your improvements.

pages/assessment-quality-dashboard-view.js
```diff
@@ -242,9 +153,89 @@ class AssessmentQualityDashboardView extends SkeletonMixin(Localizer(MobxLitElem
 	}
 
 	_renderQuizTable() {
-		this.tableConfig.heading = this.localize('dashboard:quizDetailsTitle');
+		const noStyle = styleMap();
+		const redirectToQuiz = (e) => {
```
NicholasHallman, line 157: `nit` functions that are called by an event listener should start with `handle` e.g `handleRedirectToQuiz`

pages/assessment-quality-dashboard-view.js
```diff
@@ -242,9 +153,89 @@ class AssessmentQualityDashboardView extends SkeletonMixin(Localizer(MobxLitElem
...
+						</div>
+					</div>
+			</div>`;
+			return { value: formatedValue, noStyle };
+		};
+		const averageScoreTemplate = (data) => { return { value: html`<div>${data.averageScore}</div>`, noStyle };};
+		const reliabilityCoefficientTemplate = (data) => { return  { value : html`<div>${data.reliabilityCoefficient}</div>`, noStyle };};
+		const numberOfQuizItems = (data) => { return { value: html`<div>${data.numberOfQuizItems}</div>`, noStyle };};
+		const flaggedQuizItems = (data) => { return { value: html`<div>${data.flaggedQuizItems}</div>`, noStyle };};
+		const numberOfLearnerAnswers = (data) => { return { value: html`<div>${data.numberOfLearnerAnswers}</div>`, noStyle };};
```
NicholasHallman, line 217: `nit` These lambdas don't need to live inside the render function. Might be better to place them near the top of the file outside the class since they only need to be registered once

JikaiLong:
Added new optional config` sharedStyles` in `tableConfig` which takes a html style template and apply styles to all relative elements in the table. The test case is basically checking if the shared style template exists since the` Window.getComputedStyle(query_element) `for corresponding elements always give me default style value though the shared style works visually. Hope some can tell me if it is supposed to behave like that. 
Fixed most `nit` and style-wised problem

test/components/assessment-quiz-table/assessment-quiz-table.test.js
```diff
@@ -0,0 +1,221 @@
...
+			const secondRow = el.renderRoot.querySelectorAll('d2l-table-wrapper table tbody tr')[1].querySelectorAll('td');
+			expect(secondRow[0].querySelector('div').innerText).equal('0.5');
+			expect(secondRow[1].querySelector('div').innerText).equal('0.89');
+		});
+	});
+
+	describe('render custom cell content', () => {
+		let originalOpen;
+		beforeEach(() => {
+			originalOpen = window.open;
```
Vitalii-Misechko, line 168: `nit` `originalOpen` will never change, therefore, it can be a constant.
```JS
const originalOpen = window.open;
beforeEach(() => {
```

Vitalii-Misechko (approved):
LGTM!

NicholasHallman (approved):
Looks fantastic thank you for all your work on this! 💯

PR #34, opened 2021-06-18
feat: US127982 [AQ] Quiz Details Table UI. Consume proper BDP API response

Vitalii-Misechko:
[US127982](https://rally1.rallydev.com/#/23525809118d/iterationstatus?detail=%2Fuserstory%2F603890971124&sharedViewId=328666453148&view=dc9e7043-1cf6-486d-a4aa-3a212976e789&fdp=true?fdp=true): [AQ] Quiz Details Table UI
[US128385](https://rally1.rallydev.com/#/23525809118d/iterationstatus?detail=%2Fuserstory%2F605488757484&sharedViewId=328666453148&view=dc9e7043-1cf6-486d-a4aa-3a212976e789&fdp=true?fdp=true): [AQ][Filter] Apply tree filter selection API

PR adjusts code to consume proper BDP API. It also hides `FLAGGED_QUIZ_ITEMS` column and applies number formatting.

* Functional Testing 
  * [x] Test cases
    * AQ Dashboard shows ou-tree and quiz items on Draco LMS
    * User has the ability to open filter, deselect or maintain defaults and select new filter options
    * User has the ability to clear all filters
    * User has the ability to search through filter results
* Security, devops, perf: N/A
* UX and docs
  * [ ] Demo to Kevin
  
### Screenshot

image


Vitalii-Misechko:
JFYI @Brightspace/draco-vis

model/quizData.js
```diff
@@ -43,6 +44,58 @@ class QuizOuFilterDataManager extends OuFilterDataManager {
...
+			Object.defineProperty(quiz, 'quizId', { get() { return this.Id; }, configurable: true });
+			Object.defineProperty(quiz, 'quizName', { get() { return this.Name; }, configurable: true });
+			Object.defineProperty(quiz, 'averageScore', { get() { return this.AvgScoreFrac; }, configurable: true });
+			Object.defineProperty(quiz, 'reliabilityCoefficient', { get() { return this.ReliabilityCoef; }, configurable: true });
+			Object.defineProperty(quiz, 'numberOfQuizItems', { get() { return this.NumQuizItems; }, configurable: true });
+			// TODO update when flaggedQuizItems implemented
+			Object.defineProperty(quiz, 'flaggedQuizItems', { get() { return null; }, configurable: true });
+			Object.defineProperty(quiz, 'numberOfLearnerAnswers', { get() { return this.NumLearnerAnswers; }, configurable: true });
+			Object.defineProperty(quiz, 'courseId', { get() { return this.OrgUnitId; }, configurable: true });
+			Object.defineProperty(quiz, 'courseName', { get() { return courses.has(this.OrgUnitId) ? courses.get(this.OrgUnitId).Name : null; }, configurable: true });
```
bemailloux, line 69: Interesting. What's the advantage of defining all these getters, vs just using the api-defined properties directly?

Example: why is `this._apiData.Quizzes[0].quizId` better than `this._apiData.Quizzes[0].Id`?

bemailloux (commented):
I notice any null values are just blank in the table. Are we planning on putting some placeholder there to indicate that the value is null?

image

model/quizData.js
Vitalii-Misechko: The advantage is that I don't need to refactor all existing uses of API fields, therefore, I can show changes faster (receive feedback faster). I addition, the PR is cleaner without refactoring. From the code design perspective, It's just an additional abstraction layer that provides API modification flexebility (it's like service layer DTO with field names that differ from the data layer).

> Example: why is this._apiData.Quizzes[0].quizId better than this._apiData.Quizzes[0].Id?

This is not the question that is addressed here. If this._apiData.Quizzes => serverData.quizzes, therefore, re-phrased question is _Why is serverData.quizzes[0].quizId is than serverData.quizzes[0].Id?_ The answer is that the existing code does not know about `Id` property.

Vitalii-Misechko:
> I notice any null values are just blank in the table. Are we planning on putting some placeholder there to indicate that the value is null?
> 
> image

The table shows NULLs now for `null` and `undefined` values. I've updated the screenshot in PR's details

NicholasHallman (approved):
💯

NicholasHallman:
One thing. In this design https://xd.adobe.com/view/21fb1c7a-c885-400a-a55b-887d2b26a237-74a8/screen/482a98db-ab49-49aa-99d8-efb5834825cf/ the empty cells are grey and say `no responses` Did we want to do something like that for the null and undefined data? But say `no data` instead?

Vitalii-Misechko:
> One thing. In this design https://xd.adobe.com/view/21fb1c7a-c885-400a-a55b-887d2b26a237-74a8/screen/482a98db-ab49-49aa-99d8-efb5834825cf/ the empty cells are grey and say `no responses` Did we want to do something like that for the null and undefined data? But say `no data` instead?

I believe you mention Quiz Item Detail design instead of `Quiz Details` which does not have `no responses`

bemailloux (approved):

NicholasHallman:
> I believe you mention Quiz Item Detail design instead of `Quiz Details` which does not have `no responses`

I'm more so curiouse what kevin thinks since the quiz table doesn't have any "no data" state but quiz items does. Or are you pointing out that we will never have a null or undefined case in prod and this a demo thing?

Vitalii-Misechko:
> > I believe you mention Quiz Item Detail design instead of `Quiz Details` which does not have `no responses`
> 
> I'm more so curiouse what kevin thinks since the quiz table doesn't have any "no data" state but quiz items does. 

I see. You need to know if Kevin might decide to show `no data` on this table as well. I'll ask him and get back to you on slack

> Or are you pointing out that we will never have a null or undefined case in prod and this a demo thing?

I believe it will have null or undefined.

PR #41, opened 2021-06-28
refactored quie table component accessibility

JikaiLong:
Added handlers where the user could use arrow keys to navigate the whole table, using TAB to proceed to the first cell in the next row. 

Functional Testing
All key works on both main dashboard and quizItem dashboard
Browsers (Chrome, FF, Edge)
Security, devops, perf: N/A
UX and Docs


NicholasHallman:
Noticed during the demo that when tab is pressed to select the table it selects the whole table, which prevents a screen reader user from using the ctrl+alt arrow keys from navigating initially untill they hit tab again. Could you change it so that tabbing selects the first cell immediately? Unless we are in a mobile view in which case tabbing selects the table scroll button?

JikaiLong:
> Noticed during the demo that when tab is pressed to select the table it selects the whole table, which prevents a screen reader user from using the ctrl+alt arrow keys from navigating initially untill they hit tab again. Could you change it so that tabbing selects the first cell immediately? Unless we are in a mobile view in which case tabbing selects the table scroll button and then a second tab selects the first cell?

The `d2l-table-wrapper` is where the whole table selection take place, if I remove the tab index for this one, then the table cannot be selected at all. I am not sure how the wrapper works so I may take a look into it.

NicholasHallman:
> > Noticed during the demo that when tab is pressed to select the table it selects the whole table, which prevents a screen reader user from using the ctrl+alt arrow keys from navigating initially untill they hit tab again. Could you change it so that tabbing selects the first cell immediately? Unless we are in a mobile view in which case tabbing selects the table scroll button and then a second tab selects the first cell?
> 
> The `d2l-table-wrapper` is where the whole table selection take place, if I remove the tab index for this one, then the table cannot be selected at all. I am not sure how the wrapper works so I may take a look into it.

Maybe reach out to Dave Lockhart then

JikaiLong:
Module updated. User is able to drill directly into the table cell using TAB, however the shrinking page version of the table has whole table selected to begin with and is intended.

NicholasHallman:
Is there any performance degredation if you have a bunch of table pages and pagiante to each of them? It seems like you have an event listener on each scell which might add up if twe need to add and remove them every time the table redraws.

JikaiLong:
> Is there any performance degredation if you have a bunch of table pages and pagiante to each of them? It seems like you have an event listener on each cell which might add up if twe need to add and remove them every time the table redraws. I think you should use `event delegation` here

Since there are only 100 entries per page, the performance isn't much affected when paginate to each of them. However, I did use the a global design before where I stored a global variable for current cell. But the engagement dashboard uses the same design by bind a listener for each cell, I wonder if I should be consistent or try the global way.

NicholasHallman:
> But the engagement dashboard uses the same design by bind a listener for each cell, I wonder if I should be consistent or try the global way.

Good point I thought we used event delegation more frequently but it looks like we only use it in a few places

NicholasHallman (approved):

components/table/assessment-quiz-table.js
```diff
@@ -110,9 +111,9 @@ class AssessmentTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 			</th>`;
 	}
 
-	_renderTableCell(rowItem, formatFunction) {
+	_renderTableCell(rowItem, formatFunction, columnIdx) {
 		const  { value, style } = formatFunction(rowItem);
-		return html`<td style=${style}> ${value} </td>`;
+		return html`<td style=${style} tabindex="${!this.skeleton && columnIdx === 0  ? 0 : -1}" @keydown="${this._handleCellKey}"> ${value} </td>`;
```
Vitalii-Misechko, line 116: Why do we want to add a key-down handler for each row cell? AFAIK, TAB works only for the column that contains the selection-checkbox and for the column with a link
image

In our case, only the first table row column should handle TAB.

components/table/assessment-quiz-table.js
```diff
@@ -127,11 +128,91 @@ class AssessmentTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+		} else if (e.keyCode === 39) {
+			columnNumber += 1;
+			if (columnNumber >= children.length) {
+				columnNumber = 0;
+			}
+		// press up
+		} else if (e.keyCode === 38) {
+			children = preRow;
+		// press down
+		} else if (e.keyCode === 40) {
```
Vitalii-Misechko, line 208: UP/DOWN keys work properly when a screen reader is on (tested with NVDA and Narrator). Do we need that code?

components/table/assessment-quiz-table.js
```diff
@@ -127,11 +128,91 @@ class AssessmentTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+				columnNumber = children.length - 1;
+			}
+		// press right
+		} else if (e.keyCode === 39) {
+			columnNumber += 1;
+			if (columnNumber >= children.length) {
+				columnNumber = 0;
+			}
+		// press up
+		} else if (e.keyCode === 38) {
```
Vitalii-Misechko, line 161: UP/DOWN keys work properly when a screen reader is on (tested with NVDA and Narrator). Do we need that code?

components/table/assessment-quiz-table.js
JikaiLong: 1. In the engagement dashboard table component, key-down handler is passed for each `<th>` tag and I am following that design. Nick also suggested to use `event delegation` to create one handler in the table to handle all the cell selections but I would like to keep consistent with the Engagement board for now.
2. Using TAB will only go through the first column for now but that does include the column name which does not have a clickable link, I could remove that if suggested.

components/table/assessment-quiz-table.js
NicholasHallman: For non screen reader, keyboard bound, users they don't have a way to navigate the table without the arrow key code. But I agree that we should limit it to the cells containing links on the left side

JikaiLong: This key controls are used when the screen reader is not on(say there is no mouse and you can only access table with keyboard), if it is duplicate I could remove it.

components/table/assessment-quiz-table.js
```diff
@@ -127,11 +128,91 @@ class AssessmentTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 
+	// accessbility functionality for table cells
+	_handleCellKey(e) {
+		let children = e.target.parentElement.children;
+		const rows = e.target.parentElement.parentElement.children;
+		let rowNumber = Array.from(rows).indexOf(e.target.parentElement);
+		let columnNumber = Array.from(children).indexOf(e.target);
+		const headers = e.target.parentElement.parentElement.parentElement.children[0].children[0].children;
+		// press Enter
+		if (e.keyCode === 13 && columnNumber === 0) {
```
Vitalii-Misechko, line 144: ENTER-key should work only for cells with links I guess.

components/table/assessment-quiz-table.js
```diff
@@ -127,11 +128,91 @@ class AssessmentTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 		`;
 	}
 
+	// accessbility functionality for table cells
+	_handleCellKey(e) {
+		let children = e.target.parentElement.children;
+		const rows = e.target.parentElement.parentElement.children;
+		let rowNumber = Array.from(rows).indexOf(e.target.parentElement);
+		let columnNumber = Array.from(children).indexOf(e.target);
+		const headers = e.target.parentElement.parentElement.parentElement.children[0].children[0].children;
```
Vitalii-Misechko, line 142: `arentElement.parentElement.parentElement.children[0].children[0].children` this kind of code is very frigile. It's good for POC but it's not good enough for production. Could you improve/modify it?

components/table/assessment-quiz-table.js
```diff
@@ -127,11 +128,91 @@ class AssessmentTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+		if (e.keyCode === 'Tab') e.preventDefault();
+		return false;
+	}
+
+	// accessbility functionality for headers
+	_handleHeaderKey(e) {
+		let children = e.target.parentElement.children;
+		let columnNumber = Array.from(children).indexOf(e.target);
+		const tableBody =  e.target.parentElement.parentElement.parentElement.children[1];
+		const nextRow = tableBody.children[0].children;
```
Vitalii-Misechko, line 190: Does this code crash if there are no rows? `hildren[0].children` looks like a fragile code.

components/table/assessment-quiz-table.js
```diff
@@ -127,11 +128,91 @@ class AssessmentTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+		// press TAB
+		if (e.keyCode === 'Tab') e.preventDefault();
+		return false;
+	}
+
+	// accessbility functionality for headers
+	_handleHeaderKey(e) {
+		let children = e.target.parentElement.children;
+		let columnNumber = Array.from(children).indexOf(e.target);
+		const tableBody =  e.target.parentElement.parentElement.parentElement.children[1];
```
Vitalii-Misechko, line 189: It's good for POC but it's not good enough for production. Could you improve/modify it?

components/table/assessment-quiz-table.js
NicholasHallman: Instead of referencing the children you can instead store the width and height of the table in cells and use a co-ordinate system to navigate. But again, you should really only need to keep track of the height, and move up and down along the links

components/table/assessment-quiz-table.js
JikaiLong: I could add a handler for that. Thanks for pointing it out.

components/table/assessment-quiz-table.js
JikaiLong: I think I could make some global variables for rows to get rid of these redundant tree-like searching. What do you think?

components/table/assessment-quiz-table.js
NicholasHallman: This is another advantage of event delegation because like https://github.com/Brightspace/insights-engagement-dashboard/blob/master/components/applied-filters.js you don't need to reference the parent to figure out where on the table you are focusing

components/table/assessment-quiz-table.js
Vitalii-Misechko: > 1. In the engagement dashboard table component, key-down handler is passed for each `<th>` tag and I am following that design. 

I like your approach to keeping code solutions consistent. I just don't see that Engagement Dashboard (ED) handles events from `<td>`.

> Nick also suggested to use `event delegation` to create one handler in the table to handle all the cell selections but I would like to keep consistent with the Engagement board for now.

I'm fine both ways. 

> 2. Using TAB will only go through the first column for now but that does include the column name which does not have a clickable link, I could remove that if suggested.

That's cool. My point is that `_handleCellKey` is assigned to cells that do not require any handling. 

It seems to me that you add a feature (navigation with arrow keys) that is unique to Assessment Dashboard. I don't see that feature in LMS Users table and ED tables. Have you demoed your solution to Kevin? I'm afraid that he will require that AD tables behave the same way Users table/ED tables behave.

components/table/assessment-quiz-table.js
Vitalii-Misechko: > For non screen reader, keyboard bound, users they don't have a way to navigate the table without the arrow key code. But I agree that we should limit it to the cells containing links on the left side

I assume that they should not have this ability because we don't have it in other places

components/table/assessment-quiz-table.js
JikaiLong: Sounds good. Will implement the new system.

components/table/assessment-quiz-table.js
NicholasHallman: > I assume that they should not have this ability because we don't have it in other places EDIT: (Users table, ED)
This is something we should be pursuing and if it's missing in other dashboards we should go back and add it
https://blog.d2l.dev/2021/06/30/A-Slice-of-Accessibility-Pie/

Vitalii-Misechko: That's great point. Probably, we might want these changes right in `d2l-table-wrapper`. Anyway,  I rely on Kevin's judgment/decision here. If he approves we can move further on improving the current decision.

JikaiLong:
Update: 
Just talked to Kevin and he will be investigating on how other dashboards handle the similar case. But since the Narrator mod already support Key navigation, we could put this one on hold for now and wait on further information.

JikaiLong:
close this PR for now and wait on further accessibility update.

PR #42, opened 2021-06-29
added toptips for corrsponding headers

JikaiLong:
Added tooltips for corresponding headers with content from story US129361 AC. 
Change correlation value 0 to N/A to match the description from AC.  

Functional Testing
All tooltips work and display correct content
Browsers (Chrome, FF, Edge)
Security, devops, perf: N/A
UX and Docs: N/A

@Brightspace/draco-vis


pages/quiz-item-analysis-view.js
```diff
@@ -34,7 +34,7 @@ const TOPLURE_TEMPLATE = (data) => {
 
 const TOPLURE_RATE_TEMPLATE = (data) => { return  { value : html`<div>${formatPercent(data.LureResponseRate)}</div>`, style: NO_STYLE };};
 const STD_DEV_TEMPLATE = (data) => { return { value: html`<div>${data.StdDev}</div>`, style: NO_STYLE };};
-const CORRELATION_TEMPLATE = (data) => { return { value: html`<div>${data.Correlation}</div>`, style: NO_STYLE };};
+const CORRELATION_TEMPLATE = (data) => { return { value: data.Correlation === 0 ? html`<div>N/A</div>` : html`<div>${data.Correlation}</div>`, style: NO_STYLE };};
```
NicholasHallman, line 37: Why do we return a template `<div> N/A </div>`isn't it enough to just return `'N/A'` ?

JikaiLong: Good point. I will change it right now.

pages/quiz-item-analysis-view.js
```diff
@@ -34,7 +34,7 @@ const TOPLURE_TEMPLATE = (data) => {
 
 const TOPLURE_RATE_TEMPLATE = (data) => { return  { value : html`<div>${formatPercent(data.LureResponseRate)}</div>`, style: NO_STYLE };};
 const STD_DEV_TEMPLATE = (data) => { return { value: html`<div>${data.StdDev}</div>`, style: NO_STYLE };};
-const CORRELATION_TEMPLATE = (data) => { return { value: data.Correlation === 0 ? html`<div>N/A</div>` : html`<div>${data.Correlation}</div>`, style: NO_STYLE };};
+const CORRELATION_TEMPLATE = (data) => { return { value: data.Correlation === 0 ? html`N/A` : html`<div>${data.Correlation}</div>`, style: NO_STYLE };};
```
NicholasHallman, line 37: I guess my point more so is why do we return unstylled templates at all and not just string values for all of these value, style objects? presumably what it does is
```
<td> <div> value </div> </td> 
```
Which feels unnecesary unless you are styling in the red dot next to the values right? 

We do this for discrimination, upper27, lower27, correlation etc. Seems unnecesary

JikaiLong: True. I will leave them all as string values, once the red dot story is up I will change style for the ones that require the new setup. Actually, as you pointed out, the main dashboard also uses <div>{value}</div> for most cells, I think I could change them as well in a new PR.

NicholasHallman: Yeah I think it's worth cleaning up. A side effect could be unexpected css errors if the d2l-table component changes it's query from `tr td` to `tr > td` or anything of that sort. It also takes up space in the DOM and if we have hundreds of elements in the table it could be bad for tablets

JikaiLong: all non styled template are removed, also fixed an header collapse issue caused by addition of tooltips.

NicholasHallman (approved):

Vitalii-Misechko (approved):
LGTM! Merge [visual diff PR first](https://github.com/Brightspace/insights-assessment-quality-dashboard/pull/43) before merge to master

PR #47, opened 2021-07-05
Us128659 [AQ] Tooltips for Quiz Details Table

JikaiLong:
Added new Reliability Coefficient tooltip description according to AC:
"Reliability is calculated for all questions whether the quiz questions are all dichotomous or not. Reliability will usually range from 0 to 1 with negative values being possible."
@Brightspace/draco-vis


locales/en.js
```diff
@@ -17,7 +17,7 @@ export default {
 	"dashboard:quizDetailsFlaggedQuizItems": "Flagged Quiz Items",
 	"dashboard:quizDetailsNumberOfLearnerAnswers": "Number of Learner Answers",
 	"dashboard:quizDetailsQuizId": "Quiz Id:",
-	"dashboard:quizDetailsReliabilityCoefficientTooltip": 'Reliability Coefficient',
+	"dashboard:quizDetailsReliabilityCoefficientTooltip": 'Reliability is calculated for all questions whether the quiz questions are all dichotomous or not. Reliability will usually range from 0 to 1 with negative values being possible.',
```
bemailloux, line 20: > Reliability will usually range from 0 to 1 with negative values being possible.

I think this is a very confusing explanation. In my opinion "Reliability can range from -1 to +1" is much clearer, but maybe there was a reason for the strange wording. 

Can you check with Kevin please?

bemailloux (approved):

locales/en.js
JikaiLong: sure.

bemailloux:
Why is there an empty file in the diff? 

image

Vitalii-Misechko (approved):

JikaiLong:
Kevin's Update: 
" -0.01 to -1 are equally of concern for a user, whereas 0.01, 0.5, and 1 vary in what a user would be concerned about."
 So it is better to keep the current description.

PR #53, opened 2021-07-07
added settings button on quiz item page

JikaiLong:
Functional Testing
Button works responsively and redirect to corresponding page
Browsers (Chrome, FF, Edge)
Security, devops, perf: N/A
UX and Docs: N/A
ScreenShot:
image

Immersive Nav is modified to have different back links based on where the user enters the settings page.
From quiz item dashboard: 
image

From main dashboard:
image


pages/quiz-item-analysis-view.js
```diff
@@ -327,6 +331,12 @@ class QuizItemAnalysisView extends SkeletonMixin(Localizer(MobxLitElement)) {
 		window.open(helpLink, '_blank');
 	}
 
+	_handleRedirectToSettings() {
+		const quizId = this.quizId;
+		const quizName = this.serverData.quizItemData.quizName;
+		redirect(`/settings?from=quiz,${quizId},${quizName}`);
```
NicholasHallman, line 337: Do we need the quiz name? I think we only need the quizId to return to the quiz items page

NicholasHallman (approved):

pages/quiz-item-analysis-view.js
JikaiLong: This is for the name that Immersive Nav uses that leads back to the previous page:
For example, instead of showing `Back to Quiz Item Dashboard`, it shows `Back to Poetry Term `, I could remove the naming thing if it is not important. 
image

NicholasHallman: Using it in the nav sounds good

assessment-quality-dashboard.js
```diff
@@ -77,6 +78,7 @@ class AssessmentQualityDashboard extends SkeletonMixin(Localizer(MobxLitElement)
 		if (route === '' || route === '/') return this.localize('dashboard:backToInsightsPortal');
 		else if (route.includes('/quiz')) return this.localize('dashboard:backToAssessmentDashboard');
 		else if (route.includes('/settings') && window.location.search.includes('from=main')) return this.localize('dashboard:backToAssessmentDashboard');
+		else if (route.includes('/settings') && window.location.search.includes('from=quiz')) return this.localize('dashboard:backToQuiz', { quizName: window.location.search.split(',')[2].replace('%20', ' ') });
```
bemailloux, line 81: Instead of just fixing spaces (%20) probably it would be better to use `decodeURI()`, that way all special characters would be decoded

JikaiLong: Done.

bemailloux (approved):

assessment-quality-dashboard.js
bemailloux: hmm. I wonder if that opens us up to XSS vulnerabilities though. Can you try URI-encoding some javascript and injecting it into the quiz name?

JikaiLong: Nope, since the localize takes the decoded string as a input and the code injected will not get executed.

bemailloux (approved):

PR #83, opened 2021-07-28
US130116 [AQ] Show Metron data in Quiz Items Table

Vitalii-Misechko:
[US130116](https://rally1.rallydev.com/#/23525809118d/dashboard?detail=%2Fuserstory%2F604677611411&fdp=true?fdp=true): [AQ] Show Metron data in Quiz Items Table

Related PR: https://github.com/Brightspace/bdp-query/pull/108

This PR changes:
* the code calls `/unstable/insights/data/assessmentQuality/quiz/:quizId` route
* demo data is served via msw
* response data is formatted to avoid too long numbers etc
* export data format is adjusted to what user sees on the screen

* Functional Testing 
  * [x] Test cases
    * Table loads data from `/unstable/insights/data/assessmentQuality/quiz/:quizId`
    * Page heading is the name of the quiz    
    * Table shows all nescessary columns
    * Total Questions shows valid number
    * Pagination works properly
    * ~While loading the page shows skeleton~ - for some reason `/unstable/insights/data/assessmentQuality` route gets called which leads to hiding skeleton prematurely. I'm going to address this in a separate PR
    * Export shows the same data that user sees on screen
* Security, devops, perf: N/A
* UX and docs
  * [ ] Demo to Kevin


Vitalii-Misechko:
JFYI @Brightspace/draco-vis

pages/quiz-item-analysis-view.js
```diff
@@ -180,18 +180,39 @@ class QuizItemAnalysisView extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+				if (renderFunc)	return renderFunc(data[fieldName]);
+
+				return data[fieldName];
+			};
+		};
+
+		const renderNullOrValue = (fieldName, renderFunc) => renderNullOrValueCore(fieldName, renderFunc, nullText);
+
+		const renderFuncs = [
+			renderNullOrValue('Id'),
```
NicholasHallman, line 200: It seems like all of the columns use this `renderNullOrValue` formatter, should we move it into the table so we don't have to define it for all of them here?

pages/quiz-item-analysis-view.js
```diff
@@ -180,18 +180,39 @@ class QuizItemAnalysisView extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 
 	get _dataForExport() {
-		return this._activeQuizQuestions.map(question => ([
-			question.Id,
-			question.Title,
-			formatNull(formatPercent(question.AvgScore), this.localize('dashboard:null')),
-			question.TopLure,
-			formatPercent(question.LureResponseRate),
-			formatNull(question.StdDev, this.localize('dashboard:null')),
-			formatNull(question.Correlation, this.localize('quizItem:na')),
```
Vitalii-Misechko, line 190: @JikaiLong you might notice that I've changed formatting from `N/A` to `NULL`. I've chatted with Kevin and he is okay with this change as a temporary solution. We hope to get away from `NULL`s at the end though.

NicholasHallman (approved):

pages/quiz-item-analysis-view.js
Vitalii-Misechko: Sounds reasonable. However, it contradicts the table component design when rendering/formatting happens outside the component. If we move any format-aware function into the table component it would require it to know about how to get fields' values. Therefore, we would need to pass into the component two params at least. 1st parameter that knows how to get a field value and 2nd that knows how to format that value. At this point, we just pass one render function.

Anyway, I see the possibility to address a code duplication in another approach by just declaring certain functions in the common module.

bemailloux (approved):
LGTM! You might need some changes to support https://github.com/Brightspace/bdp-query/pull/108#discussion_r678694922 though

Vitalii-Misechko:
> LGTM! You might need some changes to support [Brightspace/bdp-query#108 (comment)](https://github.com/Brightspace/bdp-query/pull/108#discussion_r678694922) though

I'll address them in separate PR if we decide to do the changes

PR #95, opened 2021-07-29
Us123996/use shared immersive nav

anhill-D2L:
Functional Testing
-keyboard navigation works
-screen reader reads it as either "Link back" on smaller screens, and "Link back to insights portal"(or similar) on larger screens
-tested on the AQD, Question analysis and quiz item pages

Performance
N/A

Security
N/A

Dev-Ops
N/A

Cost
N/A

Documentation
N/A

UX
It looks/operates the same as what we have in the engagement dashboard
Tested screen reader in Chrome
Tested keyboard navigation and responsiveness in Chrome, FF, Edge
RTL also works 

Regular size:
image

Small size:
image

RTL:
image

FYI: @Brightspace/draco-vis


NicholasHallman (approved):
Fantastic!

PR #97, opened 2021-07-29
US130116 [AQ] Show Metron data in Quiz Items Table. Show proper skeleton when loading

Vitalii-Misechko:
[US130116](https://rally1.rallydev.com/#/23525809118d/dashboard?detail=%2Fuserstory%2F604677611411&fdp=true?fdp=true): [AQ] Show Metron data in Quiz Items Table

JFYI @Brightspace/draco-vis 

Related PR: https://github.com/Brightspace/insights-assessment-quality-dashboard/pull/83

This PR ensures:
* only relevant for a page API endpoint is called (earlier quiz item page also cause a call for getting all quizzes)
* proper skeleton appears for Assessment Quality Dashboard main page
* proper skeleton appears for Quiz Item Analysis page

### Quality Notes

* Functional Testing 
  * [x] Test cases
    * The Quiz Analysis page show skeleton title when navigation was made via a bookmark
    * The Quiz Analysis page shows quiz name as title when navigation was made from AQ main page
    * The page's table is rendered as a skeleton table when data loading is in progress
    * Added unit tests that verify view of skeleton pages
* Security, devops, perf: N/A
* UX and docs
  * [x] Demo to Kevin
  
 ### Screenshots 
 
 #### Picture 1 A user will see this type of skeleton page most of the times
 
image

 #### Picture 2 A user will see this type of skeleton only the user navigates to this page via bookmark on this pagewhen 

image

  #### Picture 3
  
  
image


bemailloux:
UX idea: in my opinion it looks weird if the title is skeletonized. Maybe we should have a "static" title (like Quiz Items Analysis or something) and below it should be the skeletonized name of the quiz?

Something for Kevin to think about I guess

bemailloux (approved):
LGTM!

Vitalii-Misechko:
> UX idea: in my opinion it looks weird if the page title is skeletonized. Maybe we should have a "static" title (like Quiz Items Analysis or something) and below it should be the skeletonized name of the quiz?
> 
> Something for Kevin to think about I guess

Your idea is worth considering, however, I think we can't implement it without affecting the buttons on the right. What about an alternative based on your idea.

P.S. I'm discussing the original skeleton and the alternative with Kevin

image

pages/assessment-quality-dashboard-view.js
```diff
@@ -133,6 +133,14 @@ class AssessmentQualityDashboardView extends SkeletonMixin(Localizer(MobxLitElem
 		return itemsCount ? Math.ceil(itemsCount / this._pageSize) : 0;
 	}
 
+	async firstUpdated() {
```
NicholasHallman, line 136: Should we change this to a `connectedCallback` lifecycle function? That way the component will ask for the data before it's first render, instead of after it

route-loader.js
```diff
@@ -27,16 +26,19 @@ registerRoutes([
 		pattern: '/',
 		view: (ctx) => html`<d2l-insights-assessment-quality-dashboard-view
 							?demo=${ctx.options.isDemo}
-							.telemetry-endpoint=${ctx.options.telemetryendpoint}
-							.telemetry-id=${ctx.options.telemetryid}
+							telemetry-endpoint=${ctx.options.telemetryendpoint}
+							telemetry-id=${ctx.options.telemetryid}
```
NicholasHallman, line 30: `nit` Should `telementryEndpoint` and `telemetryId` be in camel case?

NicholasHallman (approved):
LGTM!

pages/assessment-quality-dashboard-view.js
Vitalii-Misechko: I'm hesitating about putting this code into `connectedCallback`. 
* The first render happens when properties of Lit-component have default values assigned in constructor (have not assigned with values from generated HTML). Therefore, no need to have fetched data at this point
* `connectedCallback` can be called multiple times. So, we should additionally controll on-time initialization while `firstUpdated` does it for free.

I don't have a strong opinion against `connectedCallback` but I don't have a good argument that would support it.

route-loader.js
Vitalii-Misechko: eagle eye!

pages/assessment-quality-dashboard-view.js
NicholasHallman: > The first render happens when properties of Lit-component have default values assigned in constructor (have not assigned with values from generated HTML). Therefore, no need to have fetched data at this point

Yes, the render occurs before update, which occurs after connected callback. Connected callback is the first place the html attributes are made available and so it is the first place you can use `this.metronEndpoint`. The other advantage is that the api call will be made while the child elements are constructing and running their callbacks which seems like an efficient use of cpu-time. The figure below has an a-element with a nested b-element and you can see how their lifecycles relate.

lifecycle relationship

> connectedCallback can be called multiple times. So, we should additionally controll on-time initialization while firstUpdated does it for free.

ConnectedCallback will be called again if the element is removed and re-added from the dom tree. But so would FirstUpdated. This behaviour is mitigated by the `repeat` directive

I don't think it's bad that firstUpdated is used here. I just get the impression that connectedCallback is a powerhouse lifecycle we don't use very often

Vitalii-Misechko: Thank you for this detailed comment. I'll definitely try it in the future.

Vitalii-Misechko:
@bemailloux 
> UX idea: in my opinion it looks weird if the page title is skeletonized. Maybe we should have a "static" title (like Quiz Items Analysis or something) and below it should be the skeletonized name of the quiz?
> 
> Something for Kevin to think about I guess

After discussion with Kevin we agreed to show quiz name if available (most of the time. see picture) and to show title skeleton otherwise (existing solution. it covers the case when a user make navigates to Quiz Analysis Page via bookmark )

image

bemailloux (approved):

PR #86, opened 2021-07-28
Us129568 summary container

JikaiLong:
This story also includes US129571 Add Quiz Details Card to Summary Section of Quiz Item Analysis Page.

Functional Testing

- [x] Browsers (Chrome, FF, Edge)
- [x] Responsive
- [x] Test cases
- [x] Summary Card display correct quiz detail information and components are sized properly according to AC. Skeleton page displays when the component is loading. Only max of 2 lines is allowed from each attribute and it content, otherwise text will be truncated.
- [x]  Demo to Kevin

Screenshots:
When data loading is complete
image

When data is still loading
image

@Brightspace/draco-vis

PR #103, opened 2021-08-05
Load quizzes for all pages

Vitalii-Misechko:
This PR loads all quizzes regardless they are needed or not (reverted my optimization when quizzes were not required on Quiz Analysis Page). It turns out that the Quiz Analysis page requires information that is fetched on the main page of AQD in [this PR](https://github.com/Brightspace/insights-assessment-quality-dashboard/pull/86). We probably should consider optimization of the Quiz Analysis page API calls later for the case when this page is shown bypassing the main page

* Functional Testing 
  * [x] E2E tests
    * Skeleton pages works correctly
    * Main page shows fetched data 
    * The Quiz Analysis page shows fetched data
* Security, devops, perf: N/A
* UX and docs

FYI @Brightspace/draco-vis


Vitalii-Misechko:
@JikaiLong feel free to merge this PR once it is approved

model/quizData.js
```diff
@@ -222,6 +223,10 @@ export class QuizData {
 	set currentQuizName(value) {
```
JikaiLong, line 223: The `currentQuizName` is set and passed to the Quiz Item Analysis Page through the following clicking approach: 
```
const handleRedirectToQuiz = (e) => {
			this.serverData.currentQuizName = e.target.innerText;
			redirect(`/quiz/${e.target.dataset.quizId}`);
		};
```
So it can only be set when the user enters through the main dashboard and click corresponding quiz names, this will always happen in a dev environment (dev bench always start from the main board), but if a user copies the URL and gets directly to the Quiz Item Analysis Page the title may not get set. 
Maybe we could use the `quizName` we get from the API response?

Vitalii-Misechko: Your observations are correct. 

> Maybe we could use the quizName we get from the API response?

The code does this 

image

Vitalii-Misechko: > dev bench always start from the main board

not really, you can navigate to the Quiz Item Analysis page and hit F5 then it desn't. Because we have different name coming from server.js you'll see an incorrect title on demo page

JikaiLong: Then everything LGTM.

JikaiLong (approved):

Vitalii-Misechko:
@NicholasHallman bump

JikaiLong:
I am making changes to API for optimization so this PR probably won't get merged.

Vitalii-Misechko:
> I am making changes to API for optimization so this PR probably won't get merged.

Thanks for letting me known. Closing this PR then

PR #104, opened 2021-08-05
Fix data rendering on the Quiz Analysis page

Vitalii-Misechko:
The PR targets dev branch, therefore, _does not require_ vis-diff to pass


Vitalii-Misechko:
@JikaiLong feel free to merge it if you decide it looks good

PR #122, opened 2021-08-20
US130740: [AQ] Persist filters in URL

TarasShDev:
### Functional Testing
- Unit tests pass;
- Manual testing:
  - OrgUnit filters are persisted in URL when they are changed;
  - OrgUnit filters are persisted while navigating between `AQ Dashboard page` to 
    - `Quiz item page` and back;
    - `Settings page`  and back;
  - OrgUnit filters are persisted while navigating between `Quiz item page` to
    - `Question item page` and back;
    - `Settings page`  and back;
  - `Quiz name` is correctly decoded from URL search parameters and shown in `nav back button`.
- Filters are stored when you open page by providing URL. 
### UX and docs - N/A
FYI @Brightspace/draco-vis


test/server.js
```diff
@@ -33,7 +33,7 @@ export function demoData({ selectedOrgUnitIdsCsv }) {
 		SemesterTypeId: 25,
 		isOrgUnitsTruncated: false,
 		IsQuizzesTruncated: false,
-		SelectedOrgUnitIds: selectedOrgUnitIdsCsv || []
+		SelectedOrgUnitIds: selectedOrgUnitIdsCsv?.toString().split(',').filter(x => x).map(Number) || []
```
TarasShDev, line 36: Due to the [valence api docs](https://github.com/Brightspace/valence-api-docs/pull/353/files#diff-ccea9889e264ba9dbc509af37c92bf0251963d5cf33cae3b0847cb47f10e7774R138-R142) this must be array (not string).

NicholasHallman:
I don't think we need URLState now that we have the router to manage search parameters. Specifically I think we should use RouteReactor and redirect to accomplish this https://github.com/brightspaceuilabs/router#routereactor

TarasShDev:
Hm, I used another approach without `ContextReactor` 🤔
What do you think about it?

model/quizData.js
```diff
@@ -184,11 +187,25 @@ export class QuizData {
...
+	updateOrgUnitIdsInUrl(newOrgUnitIds) {
+		const url = new URL(window.location.href);
+		if (!newOrgUnitIds || newOrgUnitIds.length === 0) {
+			url.searchParams.delete('ouf');
+			this.currentId += 1;
+			window.history.pushState({ id: this.currentId }, '', url.toString());
+		} else {
+			url.searchParams.set('ouf', newOrgUnitIds);
+			this.currentId += 1;
+			window.history.pushState({ id: this.currentId }, '', url.toString());
```
NicholasHallman, line 205: I think we can replace this with

```javascript
redirect(url.toString());
```

Same with line `201`

model/quizData.js
```diff
@@ -184,11 +187,25 @@ export class QuizData {
...
+		this.updateOrgUnitIdsInUrl(newOrgUnitIds);
 		if (this._selectorFilters.orgUnit.shouldReloadFromServer(newOrgUnitIds)) {
 			this.loadData({ newOrgUnitIds, metronEndpoint: this._metronEndpoint });
 		}
 	}
 
+	updateOrgUnitIdsInUrl(newOrgUnitIds) {
+		const url = new URL(window.location.href);
+		if (!newOrgUnitIds || newOrgUnitIds.length === 0) {
+			url.searchParams.delete('ouf');
```
NicholasHallman, line 199: I think we should make `ouf` a static getter of OrgUnitSelectorFilter so there's no guessing as to what it means down the line. Or maybe we move this function to the filter itself?

model/quizData.js
```diff
@@ -102,11 +102,14 @@ class ApiDataAdapter {
...
 export class QuizData {
+
+	static currentId = 0;
+
 	constructor({ fetchUserQuizDataFunc = fetchUserQuizData,  fetchUserQuizItemDataFunc = fetchUserQuizItemData }) {
 		this._metronEndpoint = null;
 		this._fetchUserQuizData = fetchUserQuizDataFunc;
 		this._fetchUserQuizItemData = fetchUserQuizItemDataFunc;
-		this.orgUnitTree = new Tree({});
+		this.orgUnitTree = new Tree({ selectedIds: this.orgUnitIdsFromURL });
```
NicholasHallman, line 112: It seems like with this approach we will only initialize the selection from the url when the data class is constructed on first page load. Is there any scenario where we may need to update the selection from the url without a hard refresh or re-initialising the class?

NicholasHallman: Note: not saying this is a bad thing (it's very efficient compared to checking anytime the url changes) I just want to make sure it doesn't compromise any reactive behaviour

model/quizData.js
TarasShDev: I thought we want to update the URL when we change filter. `redirect` makes the whole page reload, is it acceptable? 🤨🤔

NicholasHallman: reload doesn't refresh the page as long as the route is within the SPA path (which it is since we are only changing the search params in this case) it's a weird quirk of the router lol

model/quizData.js
TarasShDev: >Is there any scenario where we may need to update the selection from the url without a hard refresh or re-initialising the class?

I noticed only 1 scenario - when we change orgUnits in orgUnitFilter on the main page.

NicholasHallman: In that case I assume that the orgUnitFilter component would update the selection list and then push the change to the url. Because the push state / spa redirect doesn't cause a refresh the class would not be re-initialised which is good since we updated the selection list using the component instead of the url. So this all looks good to me :)

model/quizData.js
TarasShDev: Updated

model/quizData.js
TarasShDev: I updated code to use **redirect**, however I've checked on draco stack LMS, when we change selected orgUnits in `OrgUnitSelectorFilter` it redirects to the new URL which causes new API calls (attached short video). Such update ignores [these lines of code](https://github.com/Brightspace/insights-assessment-quality-dashboard/pull/122/files#diff-ee6e52dd0e4e04162a73fe68ada2c4c8b9cc7db0f2ee019ea5db725dfaa8553cR188-R190) so we fetch data in any case. 🤔

https://user-images.githubusercontent.com/46570898/133072897-1090e4e2-fca4-465c-a068-173e7a82e07c.mp4

NicholasHallman: Huh, okay then push state is good!

TarasShDev: Updated to use push state.

NicholasHallman: I think the hard refresh on redirect was because page.js was given a whole url and not a path with the search query and hash tacked on. Maybe I should create a helper for page.js that will strip the path, hash, and string params if the host is the same as the current one 🤔

NicholasHallman (approved):

model/quizData.js
```diff
@@ -222,6 +223,13 @@ export class QuizData {
 	set currentQuizName(value) {
 		this._currentQuizName = value;
 	}
+
+	get orgUnitIdsFromURL() {
+		const url = new URL(window.location.href);
+		const params = url.searchParams.get('ouf');
+		return params?.split(',').filter(Number).map(Number);
+	}
```
Vitalii-Misechko, line 231: IMHO `updateOrgUnitIdsInUrl`, `orgUnitIdsFromURL`, and `getSearchParams` doesn't really belong to the classes where they are declared. I would suggest considering a separate file like `route-helper` to gather these functions together.

pages/assessment-quality-dashboard-view.js
```diff
@@ -219,7 +219,8 @@ class AssessmentQualityDashboardView extends SkeletonMixin(Localizer(MobxLitElem
 
 		const handleRedirectToQuiz = (e) => {
 			this.serverData.currentQuizName = e.target.innerText;
-			redirect(`/quiz/${e.target.dataset.quizId}`);
+			const searchParams = window.location.search;
+			redirect(`/quiz/${e.target.dataset.quizId}${searchParams}`);
```
Vitalii-Misechko, line 223: I see that almost all calls of `redirect` have some code that manipulates search params. I would suggest considering a redirect wrapper that incapsulates search param manipulation

assessment-quality-dashboard.js
```diff
@@ -82,20 +82,40 @@ class AssessmentQualityDashboard extends Localizer(MobxLitElement) {
...
+			return this.localize('dashboard:backToQuiz', { quizName: decodeURI(quizName) });
+		}
 		else if (route.includes('/quiz')) return this.localize('dashboard:backToAssessmentDashboard');
 		else if (route.includes('/settings') && window.location.search.includes('from=main')) return this.localize('dashboard:backToAssessmentDashboard');
-		else if (route.includes('/settings') && window.location.search.includes('from=quiz')) return this.localize('dashboard:backToQuiz', { quizName: decodeURI(window.location.search.split(',')[2]) });
+		else if (route.includes('/settings') && window.location.search.includes('from=quiz')) {
+			const searchParams = new URL(window.location.href).searchParams;
+			const quizName = searchParams.get('from').split(',')[2];
+			return this.localize('dashboard:backToQuiz', { quizName: decodeURI(quizName) });
+		}
```
Vitalii-Misechko, line 118: Lines 108-110 and 115-117 can be extracted into function

Vitalii-Misechko (commented):
I've added a few minor code design suggestion

model/quizData.js
TarasShDev: Replaced them into `RouteHelper` class.

pages/assessment-quality-dashboard-view.js
TarasShDev: Added new method `doRedirect` in `RouteHelper` class.

assessment-quality-dashboard.js
TarasShDev: Extracted into **generateBackToQuizTitle()** method.

assessment-quality-dashboard.js
```diff
@@ -92,43 +94,35 @@ class AssessmentQualityDashboard extends Localizer(MobxLitElement) {
...
-	}
-
 	get navBackTitle() {
 		const route = this.currentRoute;
 		if (route === '' || route === '/') return this.localize('dashboard:backToInsightsPortal');
 		else if (route.includes('/question')) {
-			const searchParams = new URL(window.location.href).searchParams;
-			const quizName = searchParams.get('from').split(',')[1];
-			return this.localize('dashboard:backToQuiz', { quizName: decodeURI(quizName) });
+			return this.generateBackToQuizTitle(1);
```
Vitalii-Misechko, line 110: Could you come up with descriptive constants for 1 and 2?

pages/assessment-quality-dashboard-view.js
Vitalii-Misechko: It looks much better!
One more suggestion. `doRedirect` does not reference any member of `RouteHelper` therefore you can export it directly.
As result, you could write code like 
```JS
doRedirect(`/quiz/${e.target.dataset.quizId}${searchParams}`);
```

TarasShDev: It does - reference on `getSearchParams`, so extracted both of these methods.

assessment-quality-dashboard.js
TarasShDev: Yep, updated.

Vitalii-Misechko (approved):
LGTM!

PR #130, opened 2021-09-04
Us129283/average quiz score settings

bemailloux:
Adds settings for the Average Score card and table column, as well as the value that is used to determine if a quiz is "flagged" or not.

image

Quality notes
* Testing
  * [x] Functional test cases
    * List items are selectable
    * Average score flag threshold does not accept values outside of 0-100, or non-number values
    * Average score flag threshold value uses the default (70%) if input is left blank
    * It saves and closes and changes the settings when Save and Close is clicked
    * It closes without changing the settings when Cancel is clicked
  * [x] Browsers: Chrome, FF, Edgium
  * [x] Responsive layout
  * [x] A11y: screen reader, keyboard nav
* Security, perf, devops, docs: N/A
* UX
  * [x] Demo / designer approval

FYI @Brightspace/draco-vis


route-loader.js
```diff
@@ -47,9 +47,12 @@ registerRoutes([
...
 	{
 		pattern: '/settings',
-		view: () => html`<d2l-insights-settings-view
-							>
-							</d2l-insights-settings-view>`,
+		view: (ctx) => html`<d2l-insights-settings-view
+							?show-average-score-col="${ctx.options.showAverageScoreCol}"
+							?show-average-score-card="${ctx.options.showAverageScoreCard}"
+							average-score-flag-threshold="${ctx.options.averageScoreFlagThreshold}"
+							@d2l-insights-settings-close="${ctx.options.handleSettingsClose}"
```
bemailloux, line 54: @NicholasHallman Not sure if you intended handlers to be passed through the router like this, but it seems to work 👍

components/settings/quiz-details-settings-list.js
```diff
@@ -0,0 +1,158 @@
...
+									value="${this.averageScoreFlagThreshold}"
+									label="${this.localize('settings:flagAverageScoreEditLabel')}"
+									label-hidden
+									placeholder="${defaultAverageScoreFlagThreshold}"
+									@change="${this._handleAverageScoreFlagThresholdChange}"
+								></d2l-input-percent>
+							</div>
+						</div>
+					</div>
+				</d2l-list-item>
```
Vitalii-Misechko, line 114: Should we move each `d2l-list-item` into a separate function just to make `render` shorter/cleaner?

components/settings/quiz-details-settings-list.js
```diff
@@ -0,0 +1,158 @@
...
+
+	_handleAverageScoreFlagThresholdChange(event) {
+		this.averageScoreFlagThreshold = event.target.value;
+	}
+
+	get settings() {
+		return {
+			showAverageScoreCol: this.showAverageScoreCol,
+			showAverageScoreCard: this.showAverageScoreCard,
+			averageScoreFlagThreshold: this.averageScoreFlagThreshold || defaultAverageScoreFlagThreshold
```
Vitalii-Misechko, line 154: if zero is legit value then `||`-operator will always replace it with `defaultAverageScoreFlagThreshold`. Is this okay?

pages/settings-view.js
```diff
@@ -126,6 +168,25 @@ class SettingsView extends SkeletonMixin(Localizer(MobxLitElement)) {
 			</footer>
 		`;
 	}
+
+	_handleSaveAndClose() {
+		const quizDetailsSettingsList = this.shadowRoot.querySelector('d2l-insights-quiz-details-settings-list');
+
+		const settings = {
+			...quizDetailsSettingsList.settings
```
Vitalii-Misechko, line 176: nice 👍

test/components/settings/quiz-details-settings-list.test.js
```diff
@@ -0,0 +1,121 @@
...
+				const listItem = el.shadowRoot.querySelector(`d2l-list-item[key="${settingKey}"]`);
+				const otherSettings = selectionTestCases.filter(testCase => testCase !== settingKey);
+
+				listItem.setSelected(true);
+				await el.updateComplete;
+
+				expect(el.settings[settingKey]).to.be.true;
+				otherSettings.forEach(otherSetting => {
+					expect(el.settings[otherSetting]).to.be.false;
+				});
```
Vitalii-Misechko, line 82: nice 👍

Vitalii-Misechko (commented):
AC: 
> 2. Users will have ability to enable/disable Average Quiz Score card from appearing on Quiz Details page
It seems you can't achieve ac 2. because AQ doesn't have `Average Quiz Score card` yet. Should we move this AC into separate story?

assessment-quality-dashboard.js
```diff
@@ -22,14 +22,23 @@ class AssessmentQualityDashboard extends Localizer(MobxLitElement) {
...
 	}
 
 	constructor() {
 		super();
 		this.router = new RouteReactor(this);
 		this._serverData = new QuizData({});
+
+		this.showAverageScoreCol = false;
+		this.showAverageScoreCard = false;
+		this.averageScoreFlagThreshold = 70;
```
Vitalii-Misechko, line 41: Interesting [US129560](https://rally1.rallydev.com/#/?detail=/userstory/602834337163&fdp=true): _[AQ] Add Quiz Detail Table Column Config Support to Settings Page - UI_ describes 60 as the default value

bemailloux: You are right - I have no idea where I got 70% from unless the design has changed recently. I'll change this

components/settings/quiz-details-settings-list.js
bemailloux: It's not okay. Great catch!

bemailloux:
Something is wrong with screen reading - NVDA seems to randomly stop speaking in the middle of a sentence in the item descriptions. Will fix

route-loader.js
NicholasHallman: Oh that's a clever way of using it :O Nice find!

components/settings/quiz-details-settings-list.js
bemailloux: Since we need to do validation for the entire settings page, I delegated that responsibility to settings-view component. Therefore this component no longer bothers to check whether its value is valid or not.

bemailloux:
> Something is wrong with screen reading - NVDA seems to randomly stop speaking in the middle of a sentence in the item descriptions. Will fix

Figured it out - it was actually an NVDA setting. Changing it fixed the issue

assessment-quality-dashboard.js
```diff
@@ -135,6 +148,16 @@ class AssessmentQualityDashboard extends Localizer(MobxLitElement) {
 		`;
 	}
 
+	_handleSettingsClose(event) {
+		// apply settings if user selected "Save and Close"
+		if (event.detail) {
```
JikaiLong, line 153: I found that the AQ page does not pick up the data from the entry point when handling setting changes. Seems like the page is not updated along with the entry point change.

bemailloux: >Seems like the page is not updated

Can you clarify what you expect to happen here? There is a top-level property that is updated, but that update hasn't been passed down to the table component - is that what you're talking about?

JikaiLong: Yes, it would be nice if that is included here.

bemailloux: It might be better to create a new story to apply the settings, for a couple of reasons:
- The Average Quiz Score card (and other cards) aren't developed yet, so settings can't be applied
- Creating a "table cell with warning" style and making it appear when it should is non-trivial and shouldn't be in this story, in my opinion

JikaiLong: sounds good to me.

bemailloux: Thanks for mentioning it - I told Joe that a new story will be required.

bemailloux: Looks like it is captured under US129855 AC 2:
> Upon saving, the AQD UI should be updated with the relevant settings defined

assessment-quality-dashboard.js
```diff
@@ -12,6 +12,7 @@ import { MobxLitElement } from '@adobe/lit-mobx';
 import { QuizData } from './model/quizData';
 import { redirect } from '@brightspace-ui-labs/lit-router';
 import { RouteReactor } from '@brightspace-ui-labs/lit-router/RouteReactor';
+import { defaults as settingsDefaults } from './components/settings/defaults';
```
Vitalii-Misechko, line 15: nice!

components/settings/quiz-details-settings-list.js
Vitalii-Misechko: `nit` Just curious what was the reason to leave `d2l-list-item` outside `_renderAverageScoreColSettingItem`? We could do something like this

```JS
<d2l-list...>
   ${this._renderAverageScoreColSettingItem()}
   ${this._renderAverageScoreCardSettingsItem()}
</d2l-list>
```

Vitalii-Misechko (approved):
Quiz item analysis gold snapshot should contain `d2l-overflow-group`.  This particular visual diff is very flaky which is annoying. I hope to address the flakyness in my PR.

image

LGTM other than visual diff!

bemailloux:
> Just curious what was the reason to leave d2l-list-item outside _renderAverageScoreColSettingItem?

Just an aesthetic choice - in my opinion, since a "d2l-list" must have "d2l-list-items" inside it I think it makes sense to keep them together.

JikaiLong (approved):

PR #196, opened 2021-09-21
US129560 Move "High coefficients are defined as above" input to Quiz Reliability card setting

Vitalii-Misechko:
[US129560](https://rally1.rallydev.com/#/?detail=/userstory/602834337163&fdp=true): [AQ] Add Quiz Detail Table Column Config Support to Settings Page - UI
Made changes regarding the Slack [post](https://d2l.slack.com/archives/CMNGJ22G0/p1632160420116600?thread_ts=1631897075.110600&cid=CMNGJ22G0)

> Flag would stay under Reliability Coefficient table column setting as is. The high threshold input would move to the Quiz Reliability chart as it only applies to the chart. Messaging in the description would also need to change to articulate where the low threshold is defined. Any thoughts or concerns?

* Functional Testing 
  * [x] Accessibility
  * [x] RTL
  * [x] Responsive
* Security, devops, perf: N/A
* UX and docs
  * [x] Demo to Kevin
  
FYI @Brightspace/draco-vis

### Screenshot

image


bemailloux (approved):

NicholasHallman (approved):
It's confusing since both input boxes affect the chart.... Code looks good though!

Vitalii-Misechko:
> It's confusing since both input boxes affect the chart.... Code looks good though!

I agree with that. That is why Quiz Reliability setting for the chart has a bit longer description now (have a look at the new screenshot at the top of this PR)

PR #205, opened 2021-09-22
Quiz Reliability Chart Interactions & Accessibility

NicholasHallman:
## Preview

FT159Kbc0w

## Implementation Notes

FilterState replaces UrlState from the engagement dashboard. Since the router already listens to location push and pop events we use the routers ContextReactor to update the filters whenever those changes happen. The required fields are very similar to UrlState.

```javascript
class Filter extends FilterState {
  constructor() {
    super();
    this.load(FilterState.ctx) // initialize the filter with url data
  }

  get persistenceValue() { } // convert the filter datastructure to a string for storage

  set persistenceValue(value) { } // conver the url data into the filters selection datastructure
}
```

And that's it!

## Quality Notes

Testing:  

Automated  
 - [x] Should change colors when filtering
 - [x] Should change number of visible quizzes when filtering
 - [x] Should create an accessible description

Manual
 - [x] Chart reacts to changes in other filters
 - [x] Reliability Filter changes other charts
 - [x] Filter is preserved in the URL (FilterState)  
 - [x] Toasts appear
 - [x] Applied filter reads categories using a screen reader 
 - [x] Modifying the URL and navigating history update fitlers 

Documentation: N/A

Security: N/A

DevOps: N/A


NicholasHallman:
FYI: @Brightspace/draco-vis

JikaiLong (approved):
Looks good to me. That quick walk through really helps. :)

components/cards/reliability-chart.js
```diff
@@ -1,16 +1,73 @@
...
+};
+
+export class ReliabilityFilter extends CategoryFilter {
+	constructor(thresholds) {
+		super(
+			filterId,
+			'reliabilityChart:title',
+			quiz => this.selectedCategories.has(binReliability(quiz.reliabilityCoefficient, this.thresholds)),
+			'rf',
+			new Set([0, 1, 2, 3])
```
Vitalii-Misechko, line 29: `nit` could we use an object for passing these parameters for readability?

components/cards/reliability-chart.js
```diff
@@ -1,16 +1,73 @@
...
+	}
+
+	axeDescription(localizer, term) {
+		const langCategories = (['outlier', 'low', 'moderate', 'high'])
+			.filter((_, i) => this.selectedCategories.has(i))
+			.map(term => localizer(`reliabilityChart:${term}`))
+			.join(', ');
+		const localTitle = localizer('reliabilityChart:title');
+		const filterName = { chartName: localTitle, filterName: localTitle };
+		return `${localizer(term, filterName)} ${langCategories}`;
```
Vitalii-Misechko, line 54: `minor` It seems that line 54 expresses string concatenation which is not recommended for langterms. Could you turn `langCategories` into langterm parameter? 

```
"Viewing learners with {chartName} in these categories {categoryList}"
```

components/cards/reliability-chart.js
```diff
@@ -48,18 +105,45 @@ class QuizReliabilityChart extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+		};
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		this.thresholds = {
+			moderateCutoff: this.quizReliabilityFlagThreshold,
+			highCutoff: this.quizReliabilityHighThreshold
+		};
+	}
```
Vitalii-Misechko, line 122: `curious` if we turn `this.thresholds` into a getter will we need the code above (line 108 - 122)?

components/cards/reliability-chart.js
```diff
@@ -171,21 +270,11 @@ class QuizReliabilityChart extends SkeletonMixin(Localizer(MobxLitElement)) {
 	// @computed
 	get preparedData() {
 		// Temporary. These values will come from settings.
```
Vitalii-Misechko, line 272: `nit` do we need this comment?

model/categoryFilter.js
```diff
@@ -0,0 +1,73 @@
...
+		if (!isApplied) this.selectedCategories.clear();
+	}
+
+	//@action
+	setAll(allValues) {
+		this._all = allValues;
+		if (this.isAllCategoriesSelected()) {
+			this.selectedCategories.clear();
+		}
+	}
```
Vitalii-Misechko, line 29: Why do we need this method? Do we expect that list of all categories will be changed after the filter constructor? Or do we expect that list of all categories is not available at the time of calling the constructor?

model/categoryFilter.js
```diff
@@ -0,0 +1,73 @@
+import { action, computed, decorate, observable } from 'mobx';
+import { FilterState } from './filterState';
+export class CategoryFilter extends FilterState {
+	constructor(filterId, title, filter, key, all) {
+		super();
+		this.id = filterId;
+		this.title = title;
```
Vitalii-Misechko, line 7: `nit` Should we rename it into `titleTerm`?

model/categoryFilter.js
```diff
@@ -0,0 +1,73 @@
...
+		this.filter = filter;
+		this.selectedCategories = new Set();
+		this.persistenceKey = key;
+		this._all = all;
```
Vitalii-Misechko, line 11: `nit` should we name it `_allCategoriesSet`. It's not obvious what `_all` is. As an alternative, we could use comments.

model/categoryFilter.js
```diff
@@ -0,0 +1,73 @@
...
+		}
+	}
+
+	isAllCategoriesSelected() {
+
+		if (!this._all) return false;
+		const intersection = [...this._all].filter(v => this.selectedCategories.has(v));
+		if (intersection.length === this._all.size) return true;
+		return false;
+	}
```
Vitalii-Misechko, line 37: `nit` line 35-36 `return intersection.length === this._all.size;`

test/components/cards/reliability-chart.test.js
```diff
@@ -36,4 +57,69 @@ describe('Quiz Reliability Chart', () => {
...
+					console.error(givenTerm);
+					expect(term).to.not.be.undefined;
+					return;
+				}
+				if (typeof term === 'function') term = term(options);
+				return term;
+			};
+
+			expect(filter.axeDescription(localizer, 'alert:axeDescriptionRange')).to
+				.equal('Viewing learners with Quiz Reliability in these categories  Outlier, Low, High');
```
Vitalii-Misechko, line 122: I like this test. It looks neat!

Vitalii-Misechko (approved):
Have you had a chance to check if modifying URL affects filters? If yes, could you add this to quality notes?

Other than a few very minor comments looks great!

model/categoryFilter.js
NicholasHallman: This is a layover from engagement. When the user selects every bar in a chart we reset the filter so that nothing is selected. This is because the "all bars selected" and "no bars selected" behaviour is different. When no bars are selected and you pick the first bar, the other bars are disabled. But if all bars are selected and you pick a bar then that bar turns off. Visually the states are identical so we make them functionally the same too

components/cards/reliability-chart.js
NicholasHallman: Good suggestion, also if the values change the threshold isn't updated since it's in the constructor, whereas a getter would update

PR #217, opened 2021-09-29
TA157482 UI - API connection

Vitalii-Misechko:
[TA157482](https://rally1.rallydev.com/#/?detail=/task/609238188313&fdp=true): UI - API connection

* Functional Testing 
  * [x] Test cases ( I had to modify the code to get these results )
    * It shows toast with an error message when server responses other than 200 OK response. It is also verified on draco LMS
    * It throws an error if not all settings are provided
* Security, devops, perf: N/A
* UX and docs
  * [x] Demo to Kevin

* TODO
  * [x] [Update LMS property names](https://github.com/Brightspace/lms/pull/14577)
  
  ### Screenshot
image


components/settings/quiz-details-settings-list.js
```diff
@@ -323,17 +324,17 @@ class QuizDetailsSettingsList extends RtlMixin(Localizer(LitElement)) {
...
 			showQuizReliabilityCol: this.showQuizReliabilityCol,
 			showQuizReliabilityCard: this.showQuizReliabilityCard,
 			quizReliabilityFlagThreshold: this.quizReliabilityFlagThreshold,
-			quizReliabilityHighThreshold: this.quizReliabilityHighThreshold,
-
-			lowMeanThreshold: defaults.lowMeanThreshold // TODO
+			quizReliabilityHighThreshold: this.quizReliabilityHighThreshold
 		};
+
+		return Object.assign(allQuizDetailsSettings, settings);
```
bemailloux, line 337: Why not just `return settings` here? This is specifically the _quiz details_ list so it should not include settings for the quiz items page.

Vitalii-Misechko: `setting` does not contain all properties that have to be populated. Therefore, `saveSettings` function will throw the error `save settings: missing field XYZ`. 

`return settings` is the only valid code when all settings' properties are implemented. 

The alternative to `Object.assign` would be just to list all other properties as null in `settings` object

bemailloux: This is the wrong location to do `Object.assign` though, IMO. I think it should happen in `get settings` in settings-view.js, not here.

Vitalii-Misechko: `settings-view` gathers data from two settings lists. It's more a container that manages/hosts lists and their logic. Taking this into account I prefer to keep Quiz Detail logic in `quiz-details-settings-lists`. I will move the code there if you think we will benefit from such a change. Just let me know

bemailloux: Yeah but `quiz-details-settings-list` only manages half the settings. The rest of the settings will be managed in the other tab's list (which doesn't have an existing component yet).

So it makes no sense for `quiz-details-settings-list` to return all the settings, which is what you're doing here.

Vitalii-Misechko: Ohhh....I see. I put all settings into `allQuizDetailsSettings` and assumed that it contains only Quiz Details settings. I'll split `allQuizDetailsSettings` up then. Does it work for you?

bemailloux: Yup! Thanks

NicholasHallman (approved):

bbabu01 (approved):
LGTM!

PR #234, opened 2021-10-04
Us129596 high deviation card

JikaiLong:
## Implementation Notes

Created high deviation card and used better filter inheritance structure from Weak Correlation Card for Average Score Card and 
High Discrimination Card. 

## Quality Notes

Testing
- [x] Renders
- [x] Accessibility
- [x] Filter 

UX
- [x] Demo to Kevin

Documentation - N/A
DevOps - N/A
Security - N/A

image


locales/en.js
```diff
@@ -101,6 +101,8 @@ export default {
 	"quizItems:weakCorrelationDescription": "Quiz items with a correlation of {num} or lower.",
 	"quizItems:highDiscriminationCardTitle": "High Discrimination",
 	"quizItems:highDiscriminationCardMessage": "Quiz items with a wide gap between high and low performers.",
+	"quizItems:highDeviationCardTitle": "High Deviation",
+	"quizItems:highDeviationCardMessage": "Quiz {item} with deviation of {num} or greater.",
```
NicholasHallman, line 105: I don't think we need to have the `{item}` term here since the number at the front isn't necessarely part of the sentance.
For pluralized sentance
`Quiz items with deviation of 0.4 or greater.`
IMO is perfectly fine for all cases even with the 1 in front.

components/cards/average-quiz-score-card.js
```diff
@@ -1,51 +1,27 @@
...
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { defaults as settingsDefaults } from '../settings/defaults';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
 import { ToastEvent } from '@brightspace-ui-labs/toaster';
+import { ToggleFilter } from '../../model/toggleFilter';
 
 export const AVERAGE_QUIZ_SCORE_FILTER_ID = 'd2l-insights-average-quiz-score-card';
 
-export class AverageQuizScoreFilter extends FilterState {
+export class AverageQuizScoreFilter extends ToggleFilter {
```
NicholasHallman, line 13: Nice! Thanks for simplifying this

components/cards/high-discrimination-card.js
```diff
@@ -1,52 +1,28 @@
...
+import { ToggleFilter } from '../../model/toggleFilter';
 
 export const HIGH_DISCRIMINATION_FILTER_ID = 'd2l-insights-high-discrimination-card';
 
 const filterCondition = (question, threshold) => {
 	return question.DiscriminationIndex < threshold;
 };
 
-export class HighDiscriminationCardFilter extends FilterState {
+export class HighDiscriminationCardFilter extends ToggleFilter {
```
NicholasHallman, line 15: This too 🎉

locales/en.js
NicholasHallman: I'm also concerned that it's not clear how these sentances will get translated and if the resulting translations will be gramatically correct

JikaiLong: I will confirm with Kevin on this

JikaiLong: If we can achieve “plural” syntax for all cards, then we are good, but I am not sure if we can do that.

Vitalii-Misechko (approved):
LGTM!

NicholasHallman (approved):

PR #236, opened 2021-10-04
US130165 [AQ] Add Low Mean Card to Settings Page

bbabu01:
AQ Add Low Mean Card to Settings Page.

Quality Notes: Pending
- Unit tests passed

PR #240, opened 2021-10-06
High Discrimination Card Settings

NicholasHallman:
## Implementation Notes

Quiz Items Settings List

- The properties object is taken from the api example and the object keys are turned into property names
- The constructor does the same but matches those values to defaults

Settings-list-item

- Uses slots to preserve the styling of a settings list card includes slots for
- Title
- Description
- Example (card or value)
- Extras (inputs and input descriptions)

If we like these changes I can make them to the existing settings list and list items.

## Quality Notes

Testing

- [ ] Toggling high descrimination card setting hides card from dashboard
- [ ] Passes axe tests
- [ ] Constructs
- [ ] Has expected values

UX - need to demo
Docs / Security - N/A 

FYI: @Brightspace/draco-vis


NicholasHallman:
Going to wait for #236 to hit master before readying the PR so I can leverage the tests Binu wrote

PR #291, opened 2021-11-03
Split fetch api

NicholasHallman:
### Implementation Notes

This PR changes the demo mock server to mimic the new bdp-query behaviour for FQ.

### Quality Notes

Testing
 - [x] Data is retreived by issuing the post, then get command
 - [x] The Get fetch uses the path from the Post response
 - [x] Initial post contains semester query
 - [x] Subsequent Gets contain the aq hash query

Demo - N/A  

Documentation - N/A

Dev Ops
- Should we create a branch for FQ changes and merge these PR's to it untill the release is finished?

FYI: @Brightspace/draco-vis


model/dataApiClient.js
```diff
@@ -96,13 +100,19 @@ export async function fetchUserQuizData({ orgUnitIds, metronEndpoint }) {
 	}
 	const uri = url.toString();
 
-	let response = await d2lfetch.fetch(uri, { headers: { 'cache-control': 'no-store' } });
+	let response = await d2lfetch.fetch(uri, { method: 'POST', headers: { 'cache-control': 'no-store' } });
+	if (response.status !== 202) {
+		throw new Error('query-failure');
+	}
+	const rawDataPath = await response.body.getReader().read();
+	const dataPath = rawDataPath.value.reduce((arr, ascii) => modifyAndReturn(arr, () => arr.push(String.fromCharCode(ascii))), []).join('');
```
NicholasHallman, line 108: Sorry this is so gross I'm not sure why the reader returns a UInt8Array

johngwilkinson (commented):
As I mentioned on the server-side PR, we will likely be shipping AQD without flagged questions (I gather) - so this probably needs to be done as a non-breaking change, i.e. with a separate method for the new routes.

NicholasHallman:
> As I mentioned on the server-side PR, we will likely be shipping AQD without flagged questions (I gather) - so this probably needs to be done as a non-breaking change, i.e. with a separate method for the new routes.

Done, I created a switch for the class to pick which set of fetch functions to use.

model/dataApiClient.js
```diff
@@ -111,6 +118,34 @@ export async function fetchUserQuizData({ orgUnitIds, metronEndpoint }) {
...
+	if (orgUnitIds) {
+		url.searchParams.set('selectedOrgUnitIdsCsv', orgUnitIds.join(','));
+	}
+	const uri = url.toString();
+
+	let response = await d2lfetch.fetch(uri, { method: 'POST', headers: { 'cache-control': 'no-store' } });
+	if (response.status !== 202) {
+		throw new Error('query-failure');
+	}
+	const rawDataPath = await response.body.getReader().read();
```
johngwilkinson, line 132: this should be coming from the `Location` header

model/dataApiClient.js
```diff
@@ -111,6 +118,34 @@ export async function fetchUserQuizData({ orgUnitIds, metronEndpoint }) {
 	}
 }
 
+export async function fetchFQUserQuizData({ orgUnitIds, metronEndpoint }) {
+	const url = new URL(concatMetronUrl(metronEndpoint, assessmentQualityFQEndpoint));
+	if (orgUnitIds) {
+		url.searchParams.set('selectedOrgUnitIdsCsv', orgUnitIds.join(','));
```
johngwilkinson, line 124: I've asked that the API take the query as a POST body instead of query params: it's better practice now, and will be very useful later when the user settings need to be included for flag calculation.

johngwilkinson (commented):
My comments here aren't necessarily blockers depending how work proceeds - maybe you can coordinate with @JikaiLong on that

Vitalii-Misechko (approved):

NicholasHallman:
e2e aq
E2E results

model/dataApiClient.js
```diff
@@ -110,6 +110,45 @@ export async function fetchUserQuizData({ orgUnitIds, metronEndpoint }) {
 	}
 }
 
+export async function fetchFQUserQuizData({ orgUnitIds, metronEndpoint }) {
```
bemailloux, line 113: This name is kind of confusing to me, IMO it implies "there will be FQ data in the response" when that's not the case.

Maybe `fetchUserQuizDataV2` is more appropriate?

bemailloux (approved):
LGTM! Naming things is hard

NicholasHallman:
I removed the question items vis diff since it's rediculously flaky and isn't testing anything important (as was discussed in the slack)

PR #301, opened 2021-11-05
Flagging rules change 2

bemailloux:
2 changes in this PR that came out of John's questioning of how users would use these thresholds:
* Top lure response rate flag behaviour was inverted. 
  * Before: `flag if topLureResponseRate < threshold`
  * After: `flag if topLureResponseRate > threshold` -> indicates that lots of students selected the most common incorrect answer
* Discrimination index now flags if value is negative, in addition to if value is high
  * Before: `flag if discIndex > threshold`
  * After: `flag if discIndex > threshold or discIndex < 0` -> indicates that (a) high performing students are doing way better on this question than low performing students, or (b) low performing students are doing better than high performing students on this question, which is an odd situation worth flagging

NB: behaviour for discIndex column flagging is now different from High discrimination card filter. This is intentional

Quality notes
* Testing
  * Verify topLure behaviour
  * Verify discIndex behaviour
* UX
  * Get new wording for discIndex setting description from Kevin
* Devops, security, perf: N/A
* Docs
  * Maybe I should add some of the reasoning behind why the flagging is defined the way it is in code?

@Brightspace/draco-vis


bbabu01 (approved):
LGTM

PR #312, opened 2021-11-09
Fix: Filter History

NicholasHallman:
Problem:

1. Filters were being constructed anytime the page was constructed (anytime you navigated) creating a memory leak of filters that would push to the url history.
2. Filters would take turns applying their changes and then redirecting the url (saving the state).
3. Some links had `href='#'` in them which would push two histories. One handled by the browser, one handled by the router.

Solutions
1. Construct filters once on file load and give the filters the thresholds when the page is connected to the document (on redirect) to make sure the threshold don't go stale with settings.
2. Debounce the FilterState save ability so that filters group their changes before shipping them. This keeps all the newly awakened filters from spamming their state into the history. Also don't let a save occur if the components persistentState matches the url state.
3. Remove the href="#"

Quality Notes

Testing
 - Can navigate between quiz details and quiz items with one push of the back button
 - Can enable 3 filters and then disable those filters with 3 back button pushes.
 - All other tests still pass (only changed when filters receive their threshold)

UX 
 - how it should be. 
 
Docs, DevOps - N/A


johngwilkinson (approved):

pages/assessment-quality-dashboard-view.js
```diff
@@ -348,7 +357,7 @@ class AssessmentQualityDashboardView extends SkeletonMixin(Localizer(MobxLitElem
 			const formatedValue =  html`
 			<div class='d2l-table-quiz-name-header'>
 					<div class='d2l-body-compact'>
-						<d2l-link href="#" @click="${handleRedirectToQuiz}" @keypress="${handleRedirectToQuiz}" id="${quizId}" data-quiz-id="${quiz.quizId}" aria-label="${quiz.quizName}">${quiz.quizName}</d2l-link>
```
Vitalii-Misechko, line 351: `href="#"` allows Tab navigation.

model/toggleFilter.js
```diff
@@ -37,6 +36,8 @@ export class ToggleFilter extends FilterState {
 	}
 
 	set persistenceValue(value) {
+		if (this.isApplied === false && value !== '1') return;
+		if (this.isApplied === true && value === '1') return;
```
Vitalii-Misechko, line 40: `mobx` handles this for you. if isApplied is true then assigning `true` should not cause any reaction (the same is true for false value).

Vitalii-Misechko (commented):
This is a quite complex task. No wonder that I found a few bugs. Good start though!

Vitalii-Misechko:
Just one more thing to fix

1. Navigate to a quiz.
2. Activate `Low Mean` filter. I use this URL `http://localhost:8000/demo/quiz/1?hdf=&lmf=1` instead of steps 1 and 2 
3. Open the Settings page
4. Close the Setting page
5. Observe that `quizId` is lost `http://localhost:8000/demo/quiz/undefined?lmf=&hdf=`

pages/assessment-quality-dashboard-view.js
NicholasHallman: I'm going to replace it with a tabindex. I'm suprised the axe tests didn't catch this...

model/filterState.js
```diff
@@ -11,31 +11,46 @@ export class FilterState extends ContextReactor {
...
 
-	constructor() {
+	static queueChange(params) {
+		this._paramsDebounce = params;
+		clearTimeout(this._activeDebounce);
+		this._activeDebounce = setTimeout(() => {
+			const url = `${location.pathname}?${this._paramsDebounce.toString()}`;
+			redirect(url);
+		}, 100);
+	}
```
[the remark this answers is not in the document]
NicholasHallman: Fixed, Needed to change how the debouncing works. Since each filter submits it's urlSearch as a diff of the current url I needed to extract the change and batch them

components/cards/reliability-chart.js
```diff
@@ -30,16 +30,20 @@ export const binReliability = (reliabilityMetric, thresholds) => {
 };
 
 export class ReliabilityFilter extends CategoryFilter {
-	constructor(thresholds) {
+	constructor() {
```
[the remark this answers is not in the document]
NicholasHallman: Forgot to remove the memory leak for this page. fixed

johngwilkinson (approved):

pages/assessment-quality-dashboard-view.js
Vitalii-Misechko: `nit` is this selection approved by Kevin?
image

bemailloux:
> Fixed this by making sure the quiz name in the from param is url encoded. The problem was that the quiz names in the demo have a # in them

Just wondering: why are we storing the quiz name in the URL at all? Shouldn't we be able to fetch the name directly from data?

Vitalii-Misechko (approved):
LGTM!

model/filterState.js
```diff
@@ -11,31 +11,71 @@ export class FilterState extends ContextReactor {
...
 	}
 
-	constructor() {
+	// Each filter will ship it's change to the url individually
+	// This makes sure that the differences are compared to a baseline
+	// and makes all the appropriate changes
+	// E.G baseline: 001100
+	//         diff: 000100
+	//         diff: 001000
+	//        final: 000000
```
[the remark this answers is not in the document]
NicholasHallman: Is there an equizalent to flexing but in your code base? 🤔 It would make us look super smart but we would need to keep track of where the url params are in the baseline bitstring, and have some way of diffing different sized bitstring XD

NicholasHallman:
>> Fixed this by making sure the quiz name in the from param is url encoded. The problem was that the quiz names in the demo have a # in them

> Just wondering: why are we storing the quiz name in the URL at all? Shouldn't we be able to fetch the name directly from data?

That is because there is no guarentee that the quiz data will be loaded. If somebody shares a settings page url with you and the back link is to a quiz they don't have loaded we can't get the name and the immersive nav would have a hole in it.

pages/assessment-quality-dashboard-view.js
NicholasHallman: I added a style to make it an underline again on focus

bemailloux:
> If somebody shares a settings page url with you and the back link is to a quiz they don't have loaded we can't get the name and the immersive nav would have a hole in it.

I'd be tempted to have the settings page back always redirect to main to reduce complexity :shrug: anyway that's not in scope of this PR, so no action required at the moment

model/filterState.js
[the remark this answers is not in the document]
bemailloux: > diffing different sized bitstring
Leftpad :P

PR #326, opened 2021-11-17
DE46035-fix-QIA-data-reload

JikaiLong:
[DE46035](https://rally1.rallydev.com/#/?detail=/defect/614623539147&fdp=true): AQD: Quiz items: settings page causes data reload

Added a data cache to prevent QIA data from reloading upon entering or refreshing the page. 

Demo - N/A  
Documentation - N/A

@Brightspace/draco-vis


NicholasHallman (approved):
Nice! short and sweet

model/dataApiClient.js
```diff
@@ -104,32 +104,42 @@ export async function fetchUserQuizData({ orgUnitIds, metronEndpoint }) {
...
-	let response = await d2lfetch.fetch(uri, { headers: { 'cache-control': 'no-store' } });
-	let waitMs = 0;
-	while (response.status === 202) {
-		if (waitMs < 2000) waitMs += 250;
-		await new Promise(resolve => setTimeout(resolve, waitMs));
+const quizItemDataCache = new Map();
+const cacheKey = quizId => JSON.stringify(quizId || '');
 
-		response = await d2lfetch.fetch(uri, { headers: { 'cache-control': 'no-store' } });
+export async function fetchUserQuizItemData({ quizId, fullQuiz, metronEndpoint }) {
```
Vitalii-Misechko, line 116: `major` Usually, your cache key should depend on all parameters related to the business logic of a cached function. In this case `quizId` and `fullQuiz`.

model/dataApiClient.js
```diff
@@ -104,32 +104,42 @@ export async function fetchUserQuizData({ orgUnitIds, metronEndpoint }) {
...
+	const key = cacheKey(quizId);
+	if (!quizItemDataCache.has(key)) {
```
Vitalii-Misechko, line 118: `nit`  extensive use of `if-else`-statements might end up with a lot of nesting blocks and harder code to read. 

```JS
if (!quizItemDataCache.has(key)) {
    // a lot lines of code
    // a lot lines of code
    // a lot lines of code
    // a lot lines of code
} else {
    return quizItemDataCache.get(key);
}
```

can be replaced with a more linear solution. 

```JS
if (quizItemDataCache.has(key)) {
    return quizItemDataCache.get(key);
}
// a lot lines of code
// a lot lines of code
// a lot lines of code
```

Vitalii-Misechko:
- `major` There are no unit tests for this new code
- `jfyi` there is another potential solution that might fit this case and probably future cases. It's `Cache-Control` header.

model/dataApiClient.js
Vitalii-Misechko: Should you include `fullQuiz` in the `cacheKey` function?

NicholasHallman:
> jfyi there is another potential solution that might fit this case and probably future cases. It's Cache-Control header.
I think cache control is disabled here on purpose because we are relying on the server side cache to send new data only after 15 minutes. It would be nice if we could sync up those cache times though so the browser uses it's cache if the request is made before the 15 minute mark

Vitalii-Misechko:
> > jfyi there is another potential solution that might fit this case and probably future cases. It's Cache-Control header.
> 
> I think cache control is disabled here on purpose because we are relying on the server side cache to send new data only after 15 minutes. It would be nice if we could sync up those cache times though so the browser uses it's cache if the request is made before the 15 minute mark

That sounds feasible. The sever could emit `Cache-Control: max-age=` and/or `Expires` headers. No changes are required to this PR.

model/dataApiClient.js
JikaiLong: The usage for `fullQuiz` is a bit different in this case, it is only used for query AQD data if the user is enter directly from QIA page and needs to see some AQD data appear in the card, so I don't think it could be included in the `cachedKey`.

Vitalii-Misechko: Could you double-check the following case? A user post URL of QIA page and navigates to a question (`fullQuiz=true`). When it navigates back does it send `fullQuiz=false`? if true then you're right we don't need this parameter in a cache key. Otherwise, you need it.

Vitalii-Misechko: Oh... we don't need to include this flag.

test/model/dataApiClient.test.js
```diff
@@ -245,6 +245,23 @@ describe('Lms', () => {
...
+			const dataUri = 'begin:https://data.example.com/unstable/insights/data/assessmentQuality/quiz/2';
+			const expected = { the: 'data' };
+			fetchMock.post('path:/d2l/lp/auth/oauth2/token', {
+				'expires_at': Date.now() + 100000,
+				'access_token': 't'
+			});
+			fetchMock.get(dataUri, 202, { repeat: 2 });
+			fetchMock.get(dataUri, expected, { overwriteRoutes: false });
+
+			await fetchUserQuizItemData({ quizId: 2, metronEndpoint: 'https://data.example.com' });
```
Vitalii-Misechko, line 259: This test doesn't verify that data is taken from a cache. I assume that we can remove the cache and the test will continue to work.

I suggest the following implementation
```JS
const actualDataAfterFirstCall = await fetchUserQuizItemData({ quizId: 2 });
expect(actualDataAfterFirstCall ).to.deep.equal(expected);
const apiCalls = fetchMock.calls();
expect(apiCalls.length).to.equal(4);

fetchMock.resetHistory();
const actualDataAfterSecondCall = await fetchUserQuizItemData({ quizId: 2 });
expect(actualDataAfterSecondCall ).to.deep.equal(expected);
const secondTimeApiCalls = fetchMock.calls();
expect(secondTimeApiCalls length).to.equal(0);
```

Vitalii-Misechko: It also does not verify that data is taken from cache, but it verifies that API is not called for the second invocation of `fetchUserQuizItemData`. Therefore there is only one source of data left and it is the cache

JikaiLong: I am currently having this problem where `fetchMock` is still getting fetch calls, even though the cache exists. Not sure what could cause the issue in the test cases.

Vitalii-Misechko: Interesting. I've just copy-pasted the code that I've written. It works. 

image

`JFYI` the code in the comment has 3 revisions with some small fixes. I hope you use the last one

image

JikaiLong: Thank you. Let me give it a try.

JikaiLong: @Vitalii-Misechko Could you update the test cases on your side and see if the sauce test could get passed? I am not sure what is the source of the problem on my end?

Vitalii-Misechko: Sure! Other tests populate cache before this one. That is why my example works well because I ran only one test. To fix this you need to provide a unique quizId. And make one more minor change (see picture)

image

Vitalii-Misechko (approved):
LGTM! Thank you for your improvements!

PR #329, opened 2021-11-19
DE46040 QD: load url with filter specifying courses current user cannot view is quirky

Vitalii-Misechko:
[DE46040](https://rally1.rallydev.com/#/?detail=/defect/614623671889&fdp=true): AQD: load url with filter specifying courses current user cannot view is quirky

* Functional Testing 
  * [x] Test cases
     * [x] AQD filters out forbidden org unit ids from URL and does not apply them as a filter
* Security, devops, perf: N/A
* UX and docs
  * [x] Demo
  
FYI @Brightspace/draco-vis


model/quizData.js
```diff
@@ -169,6 +201,10 @@ export class QuizData {
 	}
 
 	updateOrgUnitTree() {
+		const orgUnitIdsSet = new Set(this._serverData.orgUnits.map(ou => ou.Id));
+		const selectedOrgUnitIds = this._serverData.selectedOrgUnitIds?.filter(id => orgUnitIdsSet.has(id));
```
johngwilkinson, line 205: Does this work if the user can see more than the allowed number of org units?

In this case the tree loads incrementally, and I don't 100% remember what happens with selected org units - I think maybe there's an API call specifically to get those and their ancestors - or maybe that's part of the main api call, in which I believe this will work. But I'm not sure otherwise.

Vitalii-Misechko: > Does this work if the user can see more than the allowed number of org units?

Not sure that I understand the case. What's the `allowed number of org units`?

johngwilkinson: The number at which the LMS truncates and the org unit tree loads on demand instead. 5000 by default, but adjustable using `d2l.Tools.Insights.AssessmentQuality.CourseLimit`

Vitalii-Misechko: I'll check this case

- [ ] Does this work if the user can see more than the allowed number of org units?

Vitalii-Misechko: @johngwilkinson  I've just figured out that despite the ou filter is not filtered and there is no forbidden org unit in URL. The dashboard still shows data that is filtered (the backend returns filtered data). It seems I need to modify the backend to modify the selected list before querying the quiz data.

Vitalii-Misechko: > Does this work if the user can see more than the allowed number of org units?
> 
> In this case the tree loads incrementally, and I don't 100% remember what happens with selected org units - I think maybe there's an API call specifically to get those and their ancestors - or maybe that's part of the main api call, in which I believe this will work. But I'm not sure otherwise.

Thank you @johngwilkinson for this question. It's helped me to recall the way the backend part works. `CourseLimit` is applied to selected org units this can lead to interesting cases when filtering out selected org units from `selectedOrgUnitIds` is wrong because `CourseLimit` is exceeded. Therefore, I close this PR

Vitalii-Misechko:
Related PR: https://github.com/BrightspaceUILabs/ou-filter/pull/11

PR #335, opened 2021-11-24
Fix. AQD: ctrl-click on "back to insights portal" link does not open in new tab

Vitalii-Misechko:
* Functional Testing 
  * [x] Test cases
      * [x] Ctrl-click on "Back to Insights Portal" opens Instights Portal in a new tab
      * [x] A click "Back to Insights Portal" just opne Insgihts Portal in the same tab
* Security, devops, perf: N/A
* UX and docs

* [x] Demo

  
FYI @Brightspace/draco-vis


assessment-quality-dashboard.js
```diff
@@ -23,7 +23,6 @@ class AssessmentQualityDashboard extends Localizer(MobxLitElement) {
 
 	static get properties() {
 		return {
-			isDemo: { type: Boolean, attribute: 'demo' },
```
bemailloux, line 26: Should this be deleted?

assessment-quality-dashboard.js
```diff
@@ -157,7 +156,7 @@ class AssessmentQualityDashboard extends Localizer(MobxLitElement) {
 				nav-back-link="${this.navBackLink}"
 				nav-back-title="${this.navBackTitle}"
 				nav-back-title-short="${this.navBackTitleShort}"
-				.backLinkClickHandler="${this.navRedirect}"
+				.backLinkClickHandler="${this.navBackLink.includes(insightsPortalEndpoint) ? null : this.navRedirect}"
```
bemailloux, line 159: `Checking my understanding` If handler is null, then it behaves like a normal `<a>` element. Otherwise, it uses the handler to figure out what to do?

Is there a null check in `immersive-nav` before the handler function is called?

assessment-quality-dashboard.js
Vitalii-Misechko: I believe it should be deleted. We don't have any test code in AQD

assessment-quality-dashboard.js
Vitalii-Misechko: >  If handler is null, then it behaves like a normal <a> element. Otherwise, it uses the handler to figure out what to do?

That's correct.

> Is there a null check in immersive-nav before the handler function is called?

I believe this is not handled by immersive-nav, at least I have not found the code. Anyways, it doesn't throw any errors

anhill-D2L (approved):

bemailloux (approved):

assessment-quality-dashboard.js
bemailloux: Thanks. I was confused because I couldn't figure out how demo page was showing demo data. But I figured it out 👍

PR #349, opened 2021-12-07
US131433 Load-FQ-data

JikaiLong:
### Implementation Notes

This PR implements the loading method for Flagged Questions Data from a separated fetch request. 

### Quality Notes

Testing
 - [x] Flagged Question Data will be loaded into QuizData first then added to quizzes whenever AQ data is completely loaded 
 - [x] E2E Test when FQ API is completed: AQ and FQ Data split loading, orgUnit filtering triggers new data loading

Demo - N/A  

Documentation - N/A

FYI: @Brightspace/draco-vis


model/quizData.js
```diff
@@ -149,6 +157,33 @@ export class QuizData {
 		this._serverDataProxy++;
 	}
 
+	async loadFQDataToAQData() {
+		if (this._serverData) {
+			this._serverData.quizzes.map((quiz) => {
```
NicholasHallman, line 162: map should be forEach here since we aren't returning an array, just modifying the existing one.

model/quizData.js
```diff
@@ -149,6 +157,33 @@ export class QuizData {
...
+		}
+	}
+
+	async loadFQData() {
+		this.isFQLoading = true;
+		try {
+			// remove these 2 lines with real fetching after that story is completed
+
+			// replace this line with real fetch after that story is completed
+			this._flaggedQuizData = await this._fetchFlaggedQuizData();
```
NicholasHallman, line 174: I think we should replace these functions with a real fetch, and place the example data into `test/server.js` and then mock the routes for the demo

NicholasHallman:
Looks good so far, can you add E2E testing as a functional requirement of the quality notes? Once john's changes to bdp-query are done and merged you should be able to test against draco stack

model/quizData.js
JikaiLong: Yes, I would like to wait until FQ fetch is finalized then I could get a better idea on how the real data looks like and whether a match validation is required.

model/quizData.js
```diff
@@ -138,6 +143,9 @@ export class QuizData {
 		try {
 			this._serverData  = new ApiDataAdapter(await this._fetchUserQuizData(params));
 			this.updateOrgUnitTree();
+			if (this._flaggedQuizData) {
+				await this.loadFQDataToAQData();
```
Vitalii-Misechko, line 147: `loadData` doesn't change state of `isFQlaoding`. Should it do that?

model/quizData.js
```diff
@@ -149,6 +157,30 @@ export class QuizData {
 		this._serverDataProxy++;
 	}
 
+	async loadFQDataToAQData() {
+		if (this._serverData) {
+			this._serverData.quizzes.forEach((quiz) => {
+				quiz.flaggedQuizItems = this._flaggedQuizData[quiz.quizId];
+			});
```
Vitalii-Misechko, line 164: `ApiDataAdapter` is designed as _read only_ adapter of API response payload. We could pass FQData as parameter to `ApiDataAdapter` and write a logic inside `transofrm` method to provide required field values.

pages/assessment-quality-dashboard-view.js
```diff
@@ -307,6 +307,7 @@ class AssessmentQualityDashboardView extends SortMixin(SkeletonMixin(Localizer(M
 	async firstUpdated() {
 		this.serverData.currentQuizName = null;
 		this.serverData.loadData({ metronEndpoint: this.metronEndpoint });
+		this.serverData.loadFQData();
```
Vitalii-Misechko, line 310: Can be `loadFQData` a part of `loadData`?

Vitalii-Misechko:
Added a few comments. Good start!

model/quizData.js
```diff
@@ -68,8 +72,7 @@ class ApiDataAdapter {
 			Object.defineProperty(quiz, 'averageScore', { get() { return this.AvgScoreFrac; }, configurable: true });
 			Object.defineProperty(quiz, 'reliabilityCoefficient', { get() { return this.ReliabilityCoef; }, configurable: true });
 			Object.defineProperty(quiz, 'numberOfQuizItems', { get() { return this.NumQuizItems; }, configurable: true });
-			// TODO update when flaggedQuizItems implemented
-			Object.defineProperty(quiz, 'flaggedQuizItems', { get() { return null; }, configurable: true });
+			Object.defineProperty(quiz, 'flaggedQuizItems', { set(val) {quiz._flaggedQuizItems = val;}, get() { return quiz._flaggedQuizItems; }, configurable: true });
```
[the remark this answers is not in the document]
JikaiLong: huh, not actually. My intention is to add FQ data into each quizzes when it is loaded, therefore I add a setter there for modification access. Could you provide a better solution?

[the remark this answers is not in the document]
Vitalii-Misechko: I would propose an approach when `FQData` is passed to `ApiDataAdapter`. Then `transform` would provide it via getter of `flaggedQuizItems`. I hope this feasible

model/quizData.js
```diff
@@ -138,6 +143,9 @@ export class QuizData {
 		try {
 			this._serverData  = new ApiDataAdapter(await this._fetchUserQuizData(params));
 			this.updateOrgUnitTree();
+			if (this._flaggedQuizData) {
```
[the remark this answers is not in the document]
JikaiLong: This is for the future integration once the real data structure is settled I will make change. Should probably change to empty object verification for this one.

model/quizData.js
JikaiLong: No, LoadFQDataToAQData simply add FQ data to AQ data, where we assume FQ data is loaded in its own load function.

pages/assessment-quality-dashboard-view.js
JikaiLong: it could, then `loadData ` is going to call two fetch methods with two error handler and it could be too much responsibility for just one function?

JikaiLong: The approach works like this: 
1. send out a initial POST request to boost both FQ and AQ lambda worker 
2.  keep fetching for FQ and AQ data separately with 2 GET request
3.  This step is for this story, whenever AQ data is arrived display it. If FQ data arrive first, wait for AQ data and add it to AQ data; if FQ data arrive later, just add it to AQ data.

model/quizData.js
```diff
@@ -149,6 +157,30 @@ export class QuizData {
 		this._serverDataProxy++;
 	}
 
+	async loadFQDataToAQData() {
+		if (this._serverData) {
+			this._serverData.quizzes.forEach((quiz) => {
```
johngwilkinson, line 162: You probably don't need this kind of logic, if you just provide both the quiz data and the fq data to the code rendering the various UI pieces. Initially fqData should be null, which should be rendered as loading; when it arrives, mobx will trigger the render functions again and the flag counts etc. can be shown.

johngwilkinson: To clarify: I don't think adding fqData to quiz data is a good approach. Keep fqData as a separate lookup table.

pages/assessment-quality-dashboard-view.js
Vitalii-Misechko: Thank you @JikaiLong! In this case, I'd suggest adding `withFQData` to `ApiDataAdapter` that just populates some internal fields.

```JS
class ApiDataAdapter {

	constructor(apiData, fqData) {
		this._apiData = apiData;
                this._fqData = fqData || new Map();
		this.transform();
	}
	...
	transform() {
		...
		Object.defineProperty(quiz, 'flaggedQuizItems', { get() { return this._fqData.get(quizId); }, configurable: true });
	}

	withFQData(fqData) {
	  return new ApiDataAdapter(this._apiData, fqData);
	}
	
	
	
```

model/quizData.js
JikaiLong: The major challenge I found with this separated rendering approach is to match the FQ data to corresponding AQ data. And for our current design, the table is only rendering from one data source, therefore we must modify the table, compute an additional list for FQ data to display.

model/quizData.js
```diff
@@ -56,30 +54,35 @@ class QuizOuFilterDataManager extends OuFilterDataManager {
 
 class ApiDataAdapter {
 
-	constructor(apiData) {
+	constructor(apiData, fqData) {
 		this._apiData = apiData;
+		this._fqData = fqData || new Map();
```
johngwilkinson, line 59: I believe you'll want to decorate this as a observable.

I'd recommend using a plain object `{}` instead of a Map because that's what you'll get in the API response body.

model/quizData.js
```diff
@@ -52,31 +54,35 @@ class QuizOuFilterDataManager extends OuFilterDataManager {
...
 			Object.defineProperty(quiz, 'quizName', { get() { return this.Name; }, configurable: true });
 			Object.defineProperty(quiz, 'averageScore', { get() { return this.AvgScoreFrac; }, configurable: true });
 			Object.defineProperty(quiz, 'reliabilityCoefficient', { get() { return this.ReliabilityCoef; }, configurable: true });
 			Object.defineProperty(quiz, 'numberOfQuizItems', { get() { return this.NumQuizItems; }, configurable: true });
-			// TODO update when flaggedQuizItems implemented
-			Object.defineProperty(quiz, 'flaggedQuizItems', { get() { return null; }, configurable: true });
 			Object.defineProperty(quiz, 'numberOfLearnerAnswers', { get() { return this.NumLearnerAnswers; }, configurable: true });
 			Object.defineProperty(quiz, 'courseId', { get() { return this.OrgUnitId; }, configurable: true });
 			Object.defineProperty(quiz, 'courseName', { get() { return courses.has(this.OrgUnitId) ? courses.get(this.OrgUnitId).Name : null; }, configurable: true });
+			Object.defineProperty(quiz, 'flaggedQuizItems', { get() { return fqData ? fqData[quiz.quizId] : null; }, configurable: true });
```
Vitalii-Misechko, line 77: `nit`  `fqData ?` will always be true because of this line `this._fqData = fqData || {};` in the constructor.
Therefore, `return fqData[quiz.quizId]` is sufficient. If you want to return null when `fqData[quiz.quizId]` is `undefined` then you could write `return fqData[quiz.quizId] ?? null`

model/dataApiClient.js
```diff
@@ -12,12 +12,11 @@ const assessmentQualityQuizItemsEndpoint = quizId => `unstable/insights/data/ass
 const relevantChildrenCache = new Map();
 
 function concatMetronUrl(endpoint, apiPath) {
-	let endpointClone = endpoint.split().join('');
-	if (endpoint[0] === '/') endpointClone = endpointClone.slice(1);
+	let apiPathClone = apiPath.split().join('');
+	if (apiPathClone[0] === '/') apiPathClone = apiPathClone.slice(1);
 
-	return endpoint + (endpointClone.endsWith('/') ? '' : '/') + apiPath;
+	return endpoint + (endpoint.endsWith('/') ? '' : '/') + apiPathClone;
```
NicholasHallman, line 18: Thanks for adding this :)

model/dataApiClient.js
```diff
@@ -12,12 +12,11 @@ const assessmentQualityQuizItemsEndpoint = quizId => `unstable/insights/data/ass
 const relevantChildrenCache = new Map();
 
 function concatMetronUrl(endpoint, apiPath) {
-	let endpointClone = endpoint.split().join('');
-	if (endpoint[0] === '/') endpointClone = endpointClone.slice(1);
+	let apiPathClone = apiPath.split().join('');
```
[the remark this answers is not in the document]
Vitalii-Misechko: `curious` Strings in JS are [immutable](https://developer.mozilla.org/en-US/docs/Glossary/Mutable). Therefore, I'm curious why do we need that clone operation?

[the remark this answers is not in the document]
NicholasHallman: I was cloning the string because originally I was trying to use the splice function which doesn't exist for strings (because as you said they are immutable) so we can remove this part ^. Good catch Vitalii thank you.

model/quizData.js
```diff
@@ -149,6 +166,27 @@ export class QuizData {
...
 
+	async loadFQData({ metronEndpoint = null }) {
+		this._metronEndpoint = metronEndpoint;
+		this.isLoading = true;
+		try {
+
+			this._flaggedQuizData = await this._fetchFlaggedQuizData({ response: this.useSplitDataFunction ? this._poolRes.clone() : null, metronEndpoint: metronEndpoint });
+			if (this._serverData.quizzes.length > 0) {
+				this._serverData = await this._serverData.withFQData(this._flaggedQuizData);
+			}
```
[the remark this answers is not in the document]
NicholasHallman: Agreed, `then` should be used here and it should set this.isLoadingFQ and this._serverData

model/quizData.js
```diff
@@ -193,6 +231,7 @@ export class QuizData {
 		this._routeHelper.updateOrgUnitIdsInUrl(newOrgUnitIds);
 		if (this._selectorFilters.orgUnit.shouldReloadFromServer(newOrgUnitIds)) {
 			this.loadData({ newOrgUnitIds, metronEndpoint: this._metronEndpoint });
+			this.loadFQData({ newOrgUnitIds, metronEndpoint: this._metronEndpoint });
```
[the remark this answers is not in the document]
JikaiLong: It is because `newOrgUnitIds ` is sent through the first initial POST request.

test/model/quizData.test.js
```diff
@@ -51,6 +53,13 @@ describe('QuizData', () => {
 				expect(sut.isQueryError).to.equal(true);
 				expect(sut._serverData.quizzes.length).to.equal(0);
 			});
+
+			it('should load flagged quiz data into quizzes if flagged quiz data is ready', async() => {
+				sut._flaggedQuizData = testFQData;
+				await sut.loadData({});
+				expect(sut._serverData.quizzes).to.deep.equal(QuizzesWithFQData);
```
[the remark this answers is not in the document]
JikaiLong: One reason I added `flaggedQuizItems` in the `QuizData` is for caching purpose, where if main data is not loaded, we simply save the `flaggedQuizItems` data and update it to `serverData.quizzes` once the main data is ready.

model/quizData.js
johngwilkinson, line 234: I may be missing something, but if the query parameters have changed such that a server reload is needed, the first thing that has to happen is the POST, which I think is in `startPool[Poll]`. But that doesn't seem to get called here - so new data will not be loaded, it seems.

In general: could we avoid having three separate calls and instead just have one `fetchQuizData` or similar that returns a pair of promises, or accepts two callbacks? I'm concerned that the code is getting a bit tangled with these three things happening in different places.

model/quizData.js
```diff
@@ -137,20 +138,27 @@ export class QuizData {
 		this._ouFilterDataManager = new QuizOuFilterDataManager(this);
 	}
 
+	async loadData({ newOrgUnitIds, metronEndpoint, userSettings }) {
+		await this.poolData({ newOrgUnitIds, metronEndpoint, userSettings });
```
johngwilkinson, line 142: name: `pollData`? or maybe `launchQuery`?

model/quizData.js
```diff
@@ -137,20 +138,27 @@ export class QuizData {
...
 	async poolData({ newOrgUnitIds, metronEndpoint, userSettings }) {
+		this._metronEndpoint = metronEndpoint;
+		this._userSettings = userSettings;
 		const params = {
 			orgUnitIds: newOrgUnitIds || this._selectorFilters.orgUnit.selected,
 			metronEndpoint,
 			userSettings
 		};
-		this._poolRes = await startPool(params);
+		this._urlLocations = await fetchLocationUrl(params);
```
johngwilkinson, line 155: instead of setting a field, could this be returned then passed in as a parameter to the following functions? less chance of unintended side-effects that way

test/model/quizData.test.js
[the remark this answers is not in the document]
Vitalii-Misechko: Oh ... I've missed that `serverData/ApiDataAdapter` exposes `flaggedQuizItems` property and `QuizzesWithFQData` is the _expected result_ which must include `flaggedQuizItems`. Because `server.js` contains data that returned as a response to an API call but `demoFlaggedQuizzes` is different it is transformed data. Would you mind moving it into the test file?

JikaiLong:
I am going to move the Visual Diff Test into the actual FQ column story since loading FQ data does not change anything visually.

johngwilkinson (approved):

Vitalii-Misechko:
@JikaiLong have you chance to merge master into your branch? If no then I would suggest doing that. Dave Lockhard upgraded visual-diff tests [here](https://github.com/Brightspace/insights-assessment-quality-dashboard/pull/359).

model/quizData.js
```diff
@@ -149,6 +174,25 @@ export class QuizData {
...
+			if (this._serverData.quizzes.length > 0) {
+				this._serverData = await this._serverData.withFQData(this._flaggedQuizData);
+			}
+			this.isFQLoading = false;
+
+			this.isQueryError = false;
+		} catch (ignored) {
+			console.error(ignored);
+			this._flaggedQuizData = defaultFlaggedQuizData;
+			this.isQueryError = true;
```
Vitalii-Misechko, line 190: `thoughts` There is a potential issue with `isQueryError` . Let's consider a case when `loadData` fails (sets `isQueryError == true`) but in 1 second `loadFQData` succeeds (sets `isQueryError == true`) that causing the error to be hidden.
`isQueryError` is not used anywhere so far.

pages/assessment-quality-dashboard-view.js
```diff
@@ -293,23 +294,13 @@ class AssessmentQualityDashboardView extends SortMixin(SkeletonMixin(Localizer(M
...
-
 	async firstUpdated() {
+		this.skeleton, this.skeletonFQ = true;
 		this.serverData.currentQuizName = null;
-		this.serverData.loadData({ metronEndpoint: this.metronEndpoint });
-
+		this.serverData.loadAllData({ metronEndpoint: this.metronEndpoint, userSettings: this.userSettings });
 		const disposeFn = autorun(() => {
 			this.skeleton = this.serverData.isLoading;
+			this.skeletonFQ = this.serverData.FQisLoading;
```
Vitalii-Misechko, line 303: this should be `isFQLoading`, right?

components/chart/chart.js
```diff
@@ -158,10 +158,9 @@ class Chart extends SkeletonMixin(Localizer(LitElement)) {
 					}
 				}
 			};
-
 			// Create a chart
 			H.setOptions({ lang: Object.assign({}, this._langOptions, langAccessibility) });
-			this.chart = H[constructorType](this.chartContainer, this.options, this.chartCreated.bind(this));
+			this.chart = H[constructorType](this.chartContainer, this.skeleton ? {} : this.options, this.chartCreated.bind(this));
```
Vitalii-Misechko: When `this.skeleton` is true then `d2l-insights-overlay` is brought to the front effectively hiding the chart component. We don't need that change I guess

pages/assessment-quality-dashboard-view.js
```diff
@@ -293,23 +294,13 @@ class AssessmentQualityDashboardView extends SortMixin(SkeletonMixin(Localizer(M
...
-		return this.showQuizReliabilityCard ? html`
-		<d2l-insights-reliability-chart
-			?skeleton=${this.skeleton}
-			.data=${this._filteredData}
-			quiz-reliability-flag-threshold="${this.quizReliabilityFlagThreshold}"
-			quiz-reliability-high-threshold="${this.quizReliabilityHighThreshold}"
-		></d2l-insights-reliability-chart>
-		` : html``;
-	}
-
```
Vitalii-Misechko, line 306: Why did you delete the chart?

test/server.js
```diff
@@ -176,15 +180,30 @@ const handlers = (delayMs) => [
 		}
 	}),
 
+	rest.get('https://example.com/api/metron/unstable/insights/data/assessmentQualityV2/:aq/flaggedQuestionData', (req, res, ctx) => {
+		const hash = Number(req.params.aq);
+		if (hash === queryHash) {
```
Vitalii-Misechko, line 185: I don't see where the `queryHash` is initialized. It seem this if will always be false

Vitalii-Misechko (commented):
You did great work. Almost there

NicholasHallman (approved):
LGTM thanks for making the changes :)

PR #353, opened 2021-12-09
Update quizDataV2

NicholasHallman:
Functional
 - E2E test against draco-stack
 - Compatible with new bdp AQV2
 - Compatible with old Get POLL method
 - Make demo mock match new setup
 - Update Tests

Docs - N/A
UX - N/A


Vitalii-Misechko (approved):
LGTM!

model/dataApiClient.js
```diff
@@ -124,7 +125,8 @@ export async function fetchUserQuizDataV2({ orgUnitIds, metronEndpoint }) {
 	if (response.status !== 201) {
 		throw new Error('query-failure');
 	}
-	let dataPath = await response.text();//response.headers.get('location');
+	const resultData = (await response.json())[0];
```
johngwilkinson, line 128: I'm not sure if relying on ordering is a good idea. Instead you could do `quizDataLocation = <body>.find(x => x.resultSet === 'quizData').location`. Looking ahead, it would be nice to make a little helper to do this: `function getResultSetUrl(metronUrl, body, resultSet)`. (Also, probably an async function getResultSet(...) wrapper that does the polling)

model/dataApiClient.js
```diff
@@ -135,12 +137,12 @@ export async function fetchUserQuizDataV2({ orgUnitIds, metronEndpoint }) {
...
 	const aqDataUri = (new URL(concatMetronUrl(metronEndpoint, dataPath))).toString();
 	let waitMs = 0;
-	response = await d2lfetch.fetch(`${aqDataUri}/quizData`, { headers: { 'cache-control': 'no-store' } });
+	response = await d2lfetch.fetch(`${aqDataUri}`, { method: 'GET', headers: { 'cache-control': 'no-store' } });
 	while (response.status === 202) {
 		if (waitMs < 2000) waitMs += 250;
 		await new Promise(resolve => setTimeout(resolve, waitMs));
 
-		response = await d2lfetch.fetch(`${aqDataUri}/quizData`, { headers: { 'cache-control': 'no-store' } });
+		response = await d2lfetch.fetch(`${aqDataUri}/quizData`, { method: 'GET', headers: { 'cache-control': 'no-store' } });
```
johngwilkinson, line 145: remove `/quizData` I think?

model/dataApiClient.js
```diff
@@ -110,43 +111,60 @@ export async function fetchUserQuizData({ orgUnitIds, metronEndpoint }) {
...
+		headers: { 'cache-control': 'no-store', 'content-type': 'application/json' },
+		body: JSON.stringify({
+			selectedOrgUnitIdsCsv: orgUnitIds
+		})
+	});
+	if (response.status !== 201) {
+		throw new Error('query-failure');
+	}
+
+	return await _pollResponse(response, 'quizData', metronEndpoint);
```
NicholasHallman, line 167: _pollResponse can be used to get the flagged question data as long as the assessmentQualityV2 post request returns the flaggedData location in the array

model/dataApiClient.js
```diff
@@ -111,44 +111,60 @@ export async function fetchUserQuizData({ orgUnitIds, metronEndpoint }) {
...
-	}
-	const resultData = (await response.json())[0];
-	let dataPath = resultData.location;
+const _getResultTypeLocation = async(response, resultSet, metronEndpoint) => {
+	const result = await _getResponseResultType(response, resultSet);
+	if (result === undefined) return result;
+
+	let { location } = result;
+	if (location === undefined) return location;
+	if (location[0] === '/') location = location.slice(1);
```
johngwilkinson, line 125: I wonder if `concatMetronUrl` should take care of the leading slash

johngwilkinson (approved):

model/dataApiClient.js
```diff
@@ -110,43 +111,60 @@ export async function fetchUserQuizData({ orgUnitIds, metronEndpoint }) {
...
+
+export async function fetchUserQuizDataV2({ orgUnitIds, metronEndpoint }) {
+	const url = new URL(concatMetronUrl(metronEndpoint, assessmentQualityV2Endpoint));
+	const uri = url.toString();
+
+	const response = await d2lfetch.fetch(uri, {
+		method: 'POST',
+		headers: { 'cache-control': 'no-store', 'content-type': 'application/json' },
+		body: JSON.stringify({
+			selectedOrgUnitIdsCsv: orgUnitIds
```
johngwilkinson, line 160: General question, maybe not for this PR: how easy is it to include the user's settings in here? For flagged question calcs, we will need the settings that are relevant (showQuestionMeanCol, showTopLureCol, etc.).

NicholasHallman: It should be really simple since each view and the entriypoint has access to the userSettings object

johngwilkinson (approved):

PR #359, opened 2021-12-13
switching to ESM-by-default & upgrading actions

dlockhart:
This accomplishes two things that both needed to happen together:

### 1) New Visual Diff dependencies

The main motivation for this change. By switching to the new `main` branch of the actions repo, you pick up the latest visual-diff, mocha@9 and puppeteer@12 -- this will likely case minor visual differences.

### 2) Make this project use ES Modules by default

Mocha 9 (required for the new visual-diff action) requires it, and web projects shouldn't be CommonJS anyway. I've converted the things using CommonJS things (like `require()` and `__dirname`) to their ESM equivalents.

### Other minor updates

While I was in here, I updated a few things like the latest dev-server and test-runner and linting configs.


demo/index.html
```diff
@@ -2,12 +2,10 @@
 <html lang="en" data-timezone='{"name":"Canada - Toronto","identifier":"America/Chicago"}'>
 <head>
 		<link rel="stylesheet" href="/node_modules/@brightspace-ui/core/components/demo/styles.css" type="text/css">
-		<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
```
dlockhart, line 5: No longer needed now that legacy-Edge is gone! 🎉

dlockhart:
@Vitalii-Misechko not sure if you have any ideas about what's going on with the mock service worker stuff?

It's complaining that it's not available over HTTP (since localhost doesn't use HTTPS). I'm not sure whether it's the switch to the newer `@web/test-runner` from the older Karma stuff, or the upgrade to macOS 11. I see there used be a proxy configured that mapped `/mockServiceWorker.js` to `/base/MockServiceWorker.js` but I don't believe that's needed in the new setup (it complains that it can't find the SW if I do that).

I can always back out the changes here to switch to `@web/test-runner` as that isn't needed for the visual-diff upgrade, but longer term it'd be great to get that switched as `@open-wc/testing-karma` is abandoned.

Vitalii-Misechko:
> @Vitalii-Misechko not sure if you have any ideas about what's going on with the mock service worker stuff?
> 
> It's complaining that it's not available over HTTP (since localhost doesn't use HTTPS). I'm not sure whether it's the switch to the newer `@web/test-runner` from the older Karma stuff, or the upgrade to macOS 11. I see there used be a proxy configured that mapped `/mockServiceWorker.js` to `/base/MockServiceWorker.js` but I don't believe that's needed in the new setup (it complains that it can't find the SW if I do that).
> 
> I can always back out the changes here to switch to `@web/test-runner` as that isn't needed for the visual-diff upgrade, but longer term it'd be great to get that switched as `@open-wc/testing-karma` is abandoned.

A brief investigation has shown that service workers that are used by MSW require HTTPS or localhost to be used during its works. Therefore, I assume the new webserver just exposes the test environment by using IP address.  I'll continue to work on this issue.

test/components/settings/quiz-details-settings-list.test.js
```diff
@@ -242,7 +242,7 @@ describe('d2l-insights-quiz-details-settings-list', () => {
 
 				expect(el.settings.showQuizReliabilityCard).to.be.false;
 				expect(el.settings.showQuizReliabilityCol).to.be.false;
-				expect(showQuizReliabilityColListItem.disabled).to.be.undefined;
+				expect(showQuizReliabilityColListItem.disabled).to.be.false;
```
dlockhart, line 245: I was so confused as to why this test just started failing today. Then I remembered @dbatiste updated list items yesterday to default this to `false`! 🎉

dlockhart:
Ok I've pulled out the new test-runner stuff for now since it's not an actual requirement for this PR, and I think @Vitalii-Misechko is going to look into it separately. This is ready for review.

Vitalii-Misechko (approved):
LGTM!

PR #370, opened 2021-12-24
Update spread directive to Lit 2 compatible syntax

annabelsegalld2l:
In preparation for the Lit 2 major version upgrade, Gaudi is converting all of our lit custom directives to the new syntax. The directive still functions the same for consumers, but there are some major changes as to how directives are structured. For more information, see the upgrade guide: https://lit.dev/docs/releases/upgrade/#update-custom-directive-implementations. 

I also went in and added js extensions to the imports that were missing them. In Lit 2, any pages with missing extensions on imports return a 500 error. There is no change to functionality.


model/spread.js
```diff
@@ -25,50 +27,74 @@ The type acts as a selector, dictating which properties get moved into the eleme
...
+			this.lastState[attribute] = this.target[attribute];
 		});
-		// Cache new attributes
-		stateMap.set(part, lastState);
+
+		return true;
 	}
-});
+
+	asType(target, attribute, source) {
```
annabelsegalld2l, line 58: The asType and matching functions are unchanged. The spread function is the same as well, I just moved the target, source, and lastState to be class properties.

model/spread.js
```diff
@@ -25,50 +27,74 @@ The type acts as a selector, dictating which properties get moved into the eleme
...
+	update(part, opts) {
+		const source = opts[0];
+		const types = opts.slice(1);
+		let lastState = stateMap.get(part.element);
+		if (lastState === undefined) {
+			lastState = {};
+		}
+		const state = new SpreadState(part, source, lastState);
+		const setState = state.spread(types);
+		if (setState) stateMap.set(part.element, lastState);
```
annabelsegalld2l, line 95: I also changed the spread function to return a boolean indicating if we need to store the value in the stateMap. It will only return false if a non-PropertyPart is passed as the part parameter (this should not happen but we keep it as a safeguard).

model/spread.js
```diff
@@ -25,50 +27,74 @@ The type acts as a selector, dictating which properties get moved into the eleme
...
+		const sourceAttribute = this.asType(target, attribute, source);
+		return lastAttribute === sourceAttribute;
+	}
+}
+
+class Spread extends AsyncDirective {
+	render() {
+		return noChange;
+	}
+	update(part, opts) {
```
annabelsegalld2l, line 86: Instead of accepting an arbitrary amount of parameters, the new directive syntax only takes a part (in this case a PropertyPart), and an opts array. The first element in the array is the source, and the rest of the array is the types.

model/spread.js
bemailloux, line 95: Seems weird to me that setting the weakmap entry here relies on essentially a side effect from `state.spread` (mutating the lastState object) to work correctly.

Could the map reading/writing happen inside the SpreadState class instead? That would make things a bit cleaner I think, but maybe it wouldn't work (I'm not super familiar with Lit advanced topics like this)

bemailloux: Alternatively, that migration guide says that state is stored in class instance fields. Maybe we could do away with the weakmap entirely, and use a "state" instance field instead?

dlockhart (approved):

bemailloux (approved):
Seems fine to me. I have some comments/questions but I'm not as familiar with the nuances of Lit custom directives as @NicholasHallman is.

model/spread.js
annabelsegalld2l: Good point, Bryan. The reason we need to keep the WeakMap is because the state in class instance fields is a breaking change only available in Lit 2. This approach allows us to be compatible with both Lit 1 and Lit 2 syntax. I definitely agree that getting rid of the WeakMap would clean it up once we are fully in Lit 2 😄 . For context, the lit migration will need to be done by all BSI repos at once; Gaudi is in the early stages of removing some blockers (like converting custom directives).

model/spread.js
NicholasHallman: Does changing it from a deconstruction of all properties to an array change how spread gets called? In the route-loader for example its used like...

`spread(ctx.options.userSettings, allQuizItemsMetrics)`

Where `ctx.options.userSettings` has key value pairs and `allQuizItemsMetrics` is the schematic of keys we want the element to take from the key value pairs. Would we need to change the call to

`spread([ctx.options.userSettings, allQuizItemsMetrics])` ?

annabelsegalld2l: Changing the parameter to an array actually doesn't affect the consumer calls at all. What happens on the backend is the Directive syntax moves all of the function call's parameters into the opts array. So `spread(ctx.options.userSettings, allQuizItemsMetrics)` will be mapped to the opts array `[ctx.options.userSettings, allQuizItemsMetrics]`. It's a bit confusing since the consumers call the spread() function, but the update function here does the work.

NicholasHallman (approved):

PR #390, opened 2022-01-14
DE46739 [AQ] Filters persist their state in URL

Vitalii-Misechko:
[DE46739](https://rally1.rallydev.com/#/?detail=/defect/622214852853&fdp=true): [AQ] Filters persist their state in URL

* Functional Testing 
  * [x] Test cases for each filter
    * [x] Apply a filter, navigate to settings page, navigate back and verify that filter in the same state
* Security, devops, perf: N/A
* UX and docs

* [ ] Demo
  
FYI @Brightspace/draco-vis


NicholasHallman:
So I think when I was fixing the `Need to press the back button twice before you can go back a page` bug I broke filter saving. 

Unfortunatly that bug if back again. The reason is because the filters batch up and persist their filters in the url even if the filters aren't doing anything.

NicholasHallman:
> Are you saying the bug is back? How can I reproduce it?

Yep if you click on a quiz in the table you need to hit the browser's back button twice before it will go back to the main page

Vitalii-Misechko:
> > Are you saying the bug is back? How can I reproduce it?
> 
> Yep if you click on a quiz in the table you need to hit the browser's back button twice before it will go back to the main page

Oh...I see. Now, I can reproduce it as well.

Vitalii-Misechko:
> > Are you saying the bug is back? How can I reproduce it?
> 
> Yep if you click on a quiz in the table you need to hit the browser's back button twice before it will go back to the main page

Should we address this bug in a separate PR?

NicholasHallman (approved):

Vitalii-Misechko:
bump

Vitalii-Misechko:
> > > Are you saying the bug is back? How can I reproduce it?
> > 
> > 
> > Yep if you click on a quiz in the table you need to hit the browser's back button twice before it will go back to the main page
> 
> Should we address this bug in a separate PR?

Create the defect [DE46800](https://rally1.rallydev.com/#/?detail=/defect/622656747735&fdp=true): [AD] Quiz items: clicking browser back button doesn't immediately go back to main AQ page

pages/quiz-item-analysis-view.js
```diff
@@ -342,19 +342,28 @@ class QuizItemAnalysisView extends SortMixin(SkeletonMixin(Localizer(MobxLitElem
...
 		if (!this.__filteredData) {
-			this.__filteredData = new FilteredData(this.serverData, [
-				questionMeanFilter,
-				highDeviationFilter,
-				weakCorrelationFilter,
-				highDiscriminationFilter
-			]);
+			this._filters.forEach(f => f.filter.activate());
+
+			this.__filteredData = new FilteredData(this.serverData, this._filters.map(f => f.filter));
```
bemailloux, line 358: Interesting. Why doesn't this share a `FilteredData` object with the main view?

model/filterState.js
```diff
@@ -54,6 +54,7 @@ export class FilterState extends ContextReactor {
 	}
 
 	activate() {
+		this.activated = true;
```
bemailloux, line 57: I'm missing a lot of context here since I didn't touch the AQD front-end at all. What does this do?

pages/quiz-item-analysis-view.js
Vitalii-Misechko: Despite, it's doable, there are no benefits in doing that. The main view or the dashboard shares the data fetched from the server (`serverData` instance of `QuizData`). The `FilteredData` represents some facet of that data.

Vitalii-Misechko: Each view requires its own chunk of data therefore each view has its own `FilteredData`

model/filterState.js
Vitalii-Misechko: It is used in unit tests to verify that a filter is activated after adding to `FilteredData`. 
Activation is needed to subscribe on changes in that filter by calling MobX `autorun`.

It seems I need add a comment for this field

Vitalii-Misechko: I've just added a piece of code that controls `activated` state in `FilteredData` object. I hope this makes the code better

bemailloux (approved):

Vitalii-Misechko:
I guess we should merge [this PR first](https://github.com/Brightspace/insights-assessment-quality-dashboard/pull/398).

PR #401, opened 2022-01-19
US135207 Add question pool column

bemailloux:
Adds "In Question Pool" column. Value depends on whether a QuestionPoolId is present for each question.

Corresponding back-end change: https://github.com/Brightspace/bdp-query/pull/169

image

Quality notes
* Testing
  * On local demo site and Draco LMS
    * Data: column value shows localized "Yes" when questionPoolId is present; localized "No" otherwise
    * Sorting: works; "ascending / arrow up" means that question pool q's come before non-question pool q's
    * Export: localized value is included in a new column at the end of the CSV
* Security, docs, devops, perf: N/A
* UX
  * Responsive: yes
  * Screen reader: says the value ("Yes" or "No") with NVDA
  * Demo: done

FYI @Brightspace/draco-vis


Vitalii-Misechko (approved):
LGTM!

JikaiLong (approved):

bemailloux:
> LGTM. Just a quick concern: should we display N/A to indicate if the pool is not applicable?

Given the way the back-end fetches the question pool data, the QuestionPoolId field must either contain a value that JS considers "truthy", or not exist at all. Therefore, the N/A case isn't actually possible. That tracks with the domain expectations: either a question is part of a pool or not; there isn't a "not applicable" case.

PR #406, opened 2022-01-26
DE46838: screen-reader announces when no quizzes

johngwilkinson:
FYI @Brightspace/draco-vis 

Took me quite a while to figure out that using ToastEvent is probably the "right" way to announce things in our dashboards, rather than using aria-live, role=alert, etc. directly. This way any browser quirks can be ironed out in the accessible-toaster repo instead of fixing them all over the code.

Quality Notes
- functional (via demo page and Narrator)
  - on page load, loading announcements and number of quizzes are read out
  - on filter change, filter state and number of quizzes is read out
  - on filter change with no matches remaining, the alert is read out and number of quizzes is not (in addition) read out
  - on clear filter so now there are matches, filter state and the number of quizzes is read out
- security/perf/devops: n/a
- ux
  - it's possible some screen-readers may behave differently, but that's now accessible-toaster's problem :)


pages/assessment-quality-dashboard-view.js
```diff
@@ -351,6 +352,10 @@ class AssessmentQualityDashboardView extends SortMixin(SkeletonMixin(Localizer(M
 	}
 
 	render() {
+		if (this._isDataEmpty) {
+			const description = `${this.localize('dashboard:noResultsTitle')} ${this.localize('dashboard:noResultsSubtext')}`;
+			ToastEvent.dispatch(this, null, description, false);
+		}
```
Vitalii-Misechko, line 358: works like a charm

Vitalii-Misechko (approved):

NicholasHallman (approved):
LGTM! I'm glad the toaster covers so many usecases

PR #410, opened 2022-01-27
Translations/r20.22.2

johngwilkinson:


bemailloux (approved):

locales/fr-ON.js
```diff
@@ -94,6 +99,8 @@ export default {
 	"quizItems:totalQuestions": "Nombre total de questions : {num}",
 	"quizItems:quizItemNameTemplate": '{questionName} (Id: {questionId})',
 	"quizItems:na": 'S.O.',
+	"quizItems:inQuestionPoolTrue": "Yes",
+	"quizItems:inQuestionPoolFalse": "No",
```
anhill-D2L, line 103: Just a little odd that it looks like these weren't translated here, but they were for other languages (unless if Ontario French also uses yes/no)

johngwilkinson: I think that's my mistake during merge

PR #423, opened 2022-02-04
Us135205/quiz filter

bemailloux:


test/components/quizSearchFilter.test.js
```diff
@@ -0,0 +1,55 @@
+import { expect, fixture, html, oneEvent } from '@open-wc/testing';
+import { QuizSearchFilter } from '../../components/quizSearchFilter.js';
+import { runConstructor } from '@brightspace-ui/core/tools/constructor-test-helper.js';
+
+describe('quiz search filter', () => {
```
Vitalii-Misechko, line 5: Good tests!

Vitalii-Misechko (approved):
LGTM!

Vitalii-Misechko:
`thought` should we disable user search while loading?

components/quizSearchFilter.js
```diff
@@ -0,0 +1,104 @@
...
+		this.isApplied = this._filterValue !== '';
+	}
+
+	// for UrlState
+	get persistenceKey() { return this.key; }
+
+	get persistenceValue() {
+		// known issue: trying to search for items with "&" or "+" in the title doesn't work correctly
+		// has to do with the way URLSearchParams work
+		return this._filterValue;
```
NicholasHallman, line 45: To get around this problem we should url encode the string

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

bemailloux: I tried that, doesn't work. Actually, I could use some help fixing this if you have some free time :)

pages/assessment-quality-dashboard-view.js
```diff
@@ -271,9 +274,14 @@ class AssessmentQualityDashboardView extends SortMixin(SkeletonMixin(Localizer(M
 
 	get _filteredData() {
 		if (!this.__filteredData) {
+			quizSearchFilter.activate();
```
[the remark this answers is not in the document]
bemailloux: If I just do that without any additional logic, then quiz search filter is added to the list of filters here:
image

We don't want that, according to the spec

components/quizSearchFilter.js
NicholasHallman: I'll take a look :)

components/quizSearchFilter.js
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	get filterValue() {
+		return decodeURIComponent(this._filterValue);
+	}
+
+	set filterValue(value) {
+		this._filterValue = encodeURIComponent(value) ?? '';
+		this.isApplied = this._filterValue !== '';
+	}
```
NicholasHallman, line 37: Added the encoding here to make sure that url state picks it up properly

bemailloux: Doesn't work. Try with the search string `night +` - quiz 03 should appear, but it doesn't.

Vitalii-Misechko: `encodeURIComponent(value)` when value is `null` return string `'null'`
`encodeURIComponent(value)` when value is `undefined` return string `'undefined'`

Therefore, we can't use `??` here

components/quizSearchFilter.js
```diff
@@ -0,0 +1,106 @@
...
+
+// Logic
+export class QuizSearchFilter extends FilterState {
+
+	constructor() {
+		super({ key });
+		this._filterValue = '';
+		this.isApplied = false;
+
+		this.filter = (quiz) => filterQuizByName(quiz, this.filterValue);
```
NicholasHallman, line 27: use the filterValue getter, `Night +` works now

bemailloux: Huh, that's interesting. It got double-encoded, but it works, including if the page is reloaded. Thanks for fixing this!

NicholasHallman: Welcome!

bbabu01 (approved):
LGTM

bemailloux:
> `thought` should we disable user search while loading?

Looks like this isn't necessary from a functionality perspective. Changing the filter value when the data is loading behaves the same as changing it after the data is loaded. Therefore, disabling it would be a UX improvement only. I'll see if Megan has anything to say about this

bemailloux:
> > `thought` should we disable user search while loading?
> 
> Looks like this isn't necessary from a functionality perspective. Changing the filter value when the data is loading behaves the same as changing it after the data is loaded. Therefore, disabling it would be a UX improvement only. I'll see if Megan has anything to say about this

Megan says disabling is not necessary

Vitalii-Misechko:
> > > `thought` should we disable user search while loading?
> > 
> > 
> > Looks like this isn't necessary from a functionality perspective. Changing the filter value when the data is loading behaves the same as changing it after the data is loaded. Therefore, disabling it would be a UX improvement only. I'll see if Megan has anything to say about this
> 
> Megan says disabling is not necessary

Thank you for following up on this

PR #430, opened 2022-02-14
Ta161543/filter ou filter by quiz name filter

bemailloux:
Branches in the OU filter tree are hidden if quiz search filter is hiding quizzes for those branches.

Quality notes
* Testing (test cases):
  * [x] When quiz search filter is filtered, OU filter entries are filtered correctly
  * [x] When quiz search filter is cleared, OU filter entries return to unfiltered state
  * [x] Both filters are applied if page is reloaded
  * [x] OU selection status is maintained
    * If OU is hidden by quiz search filter
    * If OU is not hidden by quiz search filter
  * [x] Only quizzes that satisfy both quiz search filter and OU filter criteria are included in results
  * [x] Filtering takes "load more" OUs into account after they are loaded
  * [x] Test on Draco LMS: Quiz search filtering and OU filter search work correctly together
* Security, docs, devops: N/A
* Perf:
  * [x] 5000 org units x 50000 quizzes: applying a quiz search filter input results in a noticeable but not long delay; approx. 1 sec or less depending on the specific input
* UX
  * [x] Demo


components/quizSearchFilter.js
```diff
@@ -22,9 +22,19 @@ export class QuizSearchFilter extends FilterState {
 	constructor() {
 		super({ key });
 		this._filterValue = '';
-		this.isApplied = false;
+		this._activated = false;
```
NicholasHallman, line 25: Does this class need it's own this._activated? Can it use this.activated from the base?

components/quizSearchFilter.js
```diff
@@ -22,9 +22,19 @@ export class QuizSearchFilter extends FilterState {
...
+		this.shouldInclude = (quiz) => filterQuizByName(quiz, this.filterValue);
+	}
+
+	activate() {
+		if (!this._activated) {
+			const url = new URL(window.location.href);
+			this._filterValue = url.searchParams.get(key) || '';
+			super.activate();
+			this._activated = true;
+		}
```
NicholasHallman, line 37: Same here. I think overriding is appropriate here but isn't it simpler to do

```javascript
activate() {
  if (!this.activated) {
    const url = new URL(window.location.href);
	this._filterValue = url.searchParams.get(key) || '';
    super.activate();
  }
}
```

model/quizData.js
```diff
@@ -160,6 +171,7 @@ export class QuizData {
 	async loadData({ metronEndpoint = null, urlLocations = null }) {
 		try {
 			this._serverData  = new ApiDataAdapter(await this._fetchUserQuizData({ locations: this.useSplitDataFunction ? urlLocations : null, metronEndpoint: metronEndpoint }), defaultFlaggedQuizData);
+			this._selectorFilters.quizName.activate();
```
NicholasHallman, line 174: Should we activate the filter anytime we load data? Does this have any side effects if there's truncation or any other need to load again?

pages/assessment-quality-dashboard-view.js
```diff
@@ -274,13 +272,8 @@ class AssessmentQualityDashboardView extends SortMixin(SkeletonMixin(Localizer(M
 
 	get _filteredData() {
 		if (!this.__filteredData) {
-			quizSearchFilter.activate();
```
NicholasHallman, line 277: I think this is better since it lives with the other filters. Did this cause a bug?

NicholasHallman (commented):
Tests look great :)

model/quizData.js
bemailloux: This just ensures the filter is activated after the first `loadData`; after that it's basically a no-op, so there are no side effects.

Vitalii-Misechko, line 174: `thoughts` It just occurred to me that I perceive this line as infiltration of presentation layer logic into the service layer. I assume it's okay by now because there are other similar cases. However, in general, I believe that page or view knows the best time when a component that affects visualization should be called/initialized/activated.

test/model/quizData.test.js
```diff
@@ -159,4 +163,62 @@ describe('QuizData', () => {
 		});
 
 	});
+
+	describe('Quiz search filter x OU filter interaction', () => {
```
Vitalii-Misechko, line 167: Thank you for these tests!

test/server.js
```diff
@@ -50,12 +50,49 @@ export function demoData({ selectedOrgUnitIdsCsv }) {
...
 
 		SemesterTypeId: 25,
-		isOrgUnitsTruncated: false,
+		IsOrgUnitsTruncated: true,
 		IsQuizzesTruncated: false,
 		SelectedOrgUnitIds: selectedOrgUnitIdsCsv?.toString().split(',').filter(x => x).map(Number) || []
 	};
 }
 
+export function perfTestData() {
```
Vitalii-Misechko, line 59: Is this function used anywhere? Should we remove it? Or should we add a comment how it can be used

pages/assessment-quality-dashboard-view.js
bemailloux: Initially I moved it since I thought it would be easier to operate on OU filter if quiz search filter also lived in QuizData. But looking at it again it seems like changing it back to the old way would work fine too. I'll try that, thanks for the suggestion!

model/quizData.js
bemailloux: similar to Nick's comment above - I'll try changing it back since it shouldn't be too difficult

pages/assessment-quality-dashboard-view.js
bemailloux: Here's the issue I bumped into when trying this, which I'd unknowingly resolved by moving QuizSearchFilter to QuizData. 

The spec is "OU filter is filtered by quiz search filter", and there are 2 cases when this happens:
- QuizSearchFilter changes. Easy to update OU filter: just tell QuizData to do that
- OU filter changes. To ensure OU filter is filtered by QuizSearchFilter, QuizData needs to get a reference to QuizSearchFilter somehow, which is awkward if QuizSearchFilter lives in the other class.

I'm thinking the cleanest solution would be to move OU filter (at least the user-facing parts of it) _out_ of QuizData and turn it into a filter for FilteredData like everything else. But that seems like a pretty heavy refactor, and keeping QuizSearchFilter in QuizData seems to work well enough for now. 

Wondering what your thoughts are? (also, @Vitalii-Misechko)

Vitalii-Misechko: I'm comfortable with the idea when the filters live inside on QuizData as long as they don't cause any side effects. Ideally, filters should be used only as predicates or an object that provides a state;

I'm not sure if it is worth doing refactoring in this PR. 
Here is an example of how I would pull out side-effects into reaction.
```JS
function applyQuizSearchToOrgUnitFilter(tree) {
		const key = 'qsf';

		if (!quizSearchFilter.filterValue) {
			tree.removeVisibilityModifier(key);
			return;
		}

		const quizzesFilteredByName = this.serverData.quizzes
			.filter(quizSearchFilter.shouldInclude);

		const coursesForVisibleQuizzes = [...new Set(quizzesFilteredByName.map(quiz => quiz.OrgUnitId))];
		if (!coursesForVisibleQuizzes.length) {
			tree.removeVisibilityModifier(key);
			return;
		}

		tree.setVisibilityModifier(key, (id) => tree.hasDescendantsInList(id, coursesForVisibleQuizzes));
	}

reaction(
  // triggers each time when orgUnitTree is changed
  () => this.serverData.orgUnitTree
  orgUnitTree => { 
    applyQuizSearchToOrgUnitFilter(orgUnitTree )
  }
);

reaction(
  // triggers each time when filterValue is changed
  () => quizSearchFilter.filterValue
  filterValue => { 
    applyQuizSearchToOrgUnitFilter(this.serverData.orgUnitTree );
  }
);

```

bemailloux: I realized there was actually no reason to defer activation - quizSearchFilter can actually just be activated in the constructor. `updateOrgUnitTree` takes care of keeping OU tree up-to-date.

Vitalii-Misechko (approved):
LGMT!

NicholasHallman (approved):

PR #433, opened 2022-02-17
Ta161544/quiz filter accessibility

bemailloux:
🍞 

also, adds a tooltip mentioning that quiz filter has an effect on OU filter.

toast

Quality notes
* Testing
  * Toast message appears when new search input is entered
  * Toast message appears when search input is cleared
  * Toast message does not appear if search input is applied or cleared repeatedly
  * Screen reader (NVDA x Chrome):
    * When focused, screen reader reads out search input label, placeholder, current value and tooltip contents
    * When input is entered, screen reader reads out new dashboard status and toast message
* Security, perf, docs, devops: N/A
* UX:
  * Accessibility: see above
  * Demo: done


locales/en.js
```diff
@@ -213,6 +213,8 @@ export default {
 	"table:sortedDescending": "Sorted Descending",
 	"table:flaggedRowToolTip": 'Action may be required.',
 
-	"quizNameFilter:label": "Search quizzes by name",
-	"quizNameFilter:placeholder": "Search by Quiz Name..."
+	"quizNameFilter:label": "Search quizzes by name.",
+	"quizNameFilter:placeholder": "Search by Quiz Name...",
+	"quizNameFilter:filterName": "Quiz Search",
+	"quizNameFilter:helpInfo": "Searching by quiz name will affect available options in the Org Unit filter."
```
bbabu01, line 219: nit: Should text 'quiz name' be in title case to match placeholder

bbabu01 (approved):
LGTM

locales/en.js
bemailloux: Maybe, I still have to check wording with Megan (this isn't something she asked for but I expect users will wonder what is going on with OU filter if we don't explicitly tell them)

Vitalii-Misechko (approved):
LGTM!

PR #440, opened 2022-02-25
fix quotes in en.js so future translations won't have issues

johngwilkinson:


Vitalii-Misechko (approved):
Thank you for the comment in PR details!

anhill-D2L (approved):

anhill-D2L:
>	'accessibility:tooltipsOn': "Column Tooltips are On",
	'accessibility:tooltipsOff': "Column Tooltips are Off",
I don't think these single quotes have a negative impact... but they just look a little out of place

PR #449, opened 2022-03-28
fix owner

dlockhart:
These need to be `Brightspace/team-name`.


johngwilkinson (approved):

johngwilkinson:
These failing visual diffs are interesting: in a different repo I'm having something similar today, where I don't think there should be a change, but it's like things are aligned just slightly differently.

dlockhart:
Yeah so those are from a very small pixel change to the height of subtle buttons [from this PR](https://github.com/BrightspaceUI/core/pull/2258). @svanherk you may want to post in `#web-platform` with a warning about this?

johngwilkinson:
> Yeah so those are from a very small pixel change to the height of subtle buttons [from this PR](https://github.com/BrightspaceUI/core/pull/2258). @svanherk you may want to post in `#web-platform` with a warning about this?

Ah, thank you!

svanherk:
>  @svanherk you may want to post in #web-platform with a warning about this?

Good call, will do!

